PR #1, opened 2019-10-13
Creating golang bindings

Asalle:
Regarding your comments to pr #269
> For Address, Context, Recevier, and Sender replace free functions with methods

In the last commit I showcased an example of how it would look like, if you find it acceptable I will write bindings to all other functions.

> For argument that are single values replace slices with passing by value or passing by pointer (depending on whether the function modifies the argument):

Yes, we can tell c-for-go to do that. Please confirm if I put the pointers in the right places.

> Replace int32 return codes with Go error

I wrote a simple converter in `roc-wrappers.go`, also I opened an issue in `c-for-go` to find out if we can generate such things in the future. Please feel free to correct me/comment the [issue](https://github.com/xlab/c-for-go/issues/75).

> Create convenient wrappers for frames of different types, as described in the issue:

TBD

> Replace LogHandler function with a Logger interface

TBD

> Allocations

TBD.

General comments:
1. Thank you so much for creating a new repo, it not only makes more sense, but also makes things much clearer to me.
2. But at the same time, now we have two sources of truth for the headers. Do you have a backward compatibility promise that will absolutely guarantee that headers will not change? If not, maybe we can think about using submodules or something like that inthe future.
3. Please don't inspect commits one by one, instead look at the code. They need to be squashed, but before I get a green light from you I'll keep them as is for simplicity. 
4. Readme is highly unproffesional and will be fixed before merge of this pr. Consider it "my" personal notes.


dshil:
Hi @Asalle, I followed your conversation with @gavv in https://github.com/roc-project/roc/pull/269. Consider comments regarding to the README as small recommendations.

roc.yml
```diff
@@ -0,0 +1,69 @@
...
+  FlagGroups:
+    - {name: "CFLAGS", traits: ["linux"], flags: [
+        "-I\"../include\"",
+    ]}
+    - {name: "LDFLAGS", traits: ["linux"], flags: [
+      "-lroc",
+    ]} 
+    
+PARSER:
+  IncludePaths: ["include", "/usr/lib/gcc/x86_64-redhat-linux/9/include"]
```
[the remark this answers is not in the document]
xlab: This line is mostly needed to have types resolved upon generation, later all work for finding correct typedefs is on Go's CGO shoulders.

roc-wrapper.go
```diff
@@ -0,0 +1,15 @@
...
+
+func convertErr(err uint32, errstr string) error {
+	if err == 0 {
+		return nil
+	}
+	return fmt.Errorf(errstr)
+}
+
+// Init parses the `ip` and `port` and initializes the Address object
+func (a *Address) Init(family Family, ip string, port uint32) error {
```
gavv: I suggest to replace this with:

```
func NewAddress(family Family, ip string, port uint32) (*Address, error)
```

It would be more go-way. I also suggest to use the same approach for other objects, e.g.:

```
func OpenSender(...) (*Sender, error)
func (sender *Sender) Bind(...) error
...
```

Also, IIRC, Go stdlib usually use `int` for ports, so we can follow the convention and replace `port uint32` with `port int` here.

roc.yml
```diff
@@ -0,0 +1,69 @@
...
+    - ../include/roc/platform.h
+    - ../include/roc/receiver.h
+    - ../include/roc/sender.h
+  FlagGroups:
+    - {name: "CFLAGS", traits: ["linux"], flags: [
+        "-I\"../include\"",
+    ]}
+    - {name: "LDFLAGS", traits: ["linux"], flags: [
+      "-lroc",
+    ]}
```
gavv: Could you explain why do we need `traits: ["linux"]` here? AFAIK cgo always uses gcc or gcc-like compilers so these options should work everywhere?

roc/types.go
```diff
@@ -0,0 +1,91 @@
...
+// Address as declared in roc/address.h:59
+type Address struct {
+	ref94efe9d7    *C.roc_address
+	allocs94efe9d7 interface{}
+}
+
+// ContextConfig as declared in roc/config.h:147
+type ContextConfig struct {
+	maxPacketSize  uint32
+	maxFrameSize   uint32
```
gavv: All fields in ContextConfig, SenderConfig, and ReceiverConfig are unexported. We should export them I guess?

roc/types.go
```diff
@@ -0,0 +1,91 @@
...
+	allocs9cfc9e50          interface{}
+}
+
+// Context as declared in roc/context.h:41
+type Context C.roc_context
+
+// Frame as declared in roc/frame.h:42
+type Frame struct {
+	samples       unsafe.Pointer
+	samplesSize   uint
```
gavv: As for the `Frame` struct, I suggest to completely hide this struct from user and provide type-safe wrappers instead (https://github.com/roc-project/roc/issues/248).

roc/roc.go
```diff
@@ -1,3 +1,169 @@
...
+// AddressFamily function as declared in roc/address.h:97
+func AddressFamily(address *Address) Family {
+	caddress, _ := address.PassRef()
+	__ret := C.roc_address_family(caddress)
+	__v := (Family)(__ret)
+	return __v
+}
+
+// AddressIp function as declared in roc/address.h:116
+func AddressIp(address *Address, buf *byte, bufsz uint) string {
```
gavv: This is probably incorrect. `buf` should be a pointer to a buffer of size `bufsz`, and here we're passing a pointer to a single byte. Shouldn't we pass a slice instead?

BTW, a high level wrapper for this function may have much simpler signature:

```
func (addr *Address) IP() string
```

gavv:
> In the last commit I showcased an example of how it would look like, if you find it acceptable I will write bindings to all other functions.

Almost good. See https://github.com/roc-project/roc-go/pull/1#discussion_r334563360

> Yes, we can tell c-for-go to do that. Please confirm if I put the pointers in the right places.

Almost. See https://github.com/roc-project/roc-go/pull/1#discussion_r334568614

README.md
```diff
@@ -6,3 +6,14 @@ Go bindings for Roc
 ```
 go get github.com/roc-project/roc-go/roc
 ```
+
+# Build
+1. Clone roc: `git clone git@github.com:roc-project/roc.git`
+2. Install SCons: `sudo dnf install scons`
+3. Install some other roc dependencies when prompted by SCons, but not `openfec` (I couldn't find an easy way to install it on fedora).
+4. Build and install roc and its dependency (openfec): `sudo scons . --build-3rdparty=openfec --disable-tests install`
+5. Now that libroc is installed system-wide, goto roc-go repo and gen the go bindings: `c-for-go ./roc.yml`
```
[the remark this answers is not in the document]
xlab: I'd not expect that bindings generation would work on any platform, it is supposed to work only on dev's machine. It's very optimistic to generate bindings dynamically upon installation.
I usually generate bindings that are common for each platform (using UNIX env on my macOS), and then add required differencies for win, pure linux, etc.

roc/types.go
xlab: Types like this one (everything in `types.go`) are autogenerated from C header. Depending on the lib it's often required to expose everything as-is and say that this lib is "low level". As it  follows its C API that was not uder-friendly in the first place. Or, alternatively, to have a low-level generated bindings aside in private package and expose nice and clean types and methods suitable for end-user.

Example: https://github.com/xlab/pocketsphinx-go (there are nice and clean wrappers, that make calls into autogenerated Go bindings).

roc.yml
xlab: These are platform dependent, even for FreeBSD/Ubuntu the set of libraries might have different names in it. Not sure if true for roc, though.

gavv:
> But at the same time, now we have two sources of truth for the headers.

Yes. But there would be two sources anyway: one copy of headers used when building libroc (by user or e.g. by deb-package maintainer) and another copy used when building bindings (by go get).

Why? Because when the user calls go get, it will create its own git checkout. We can't force it to use our checkout (and it would break the usual user experience with go get).

So even if headers are stored only in a single repo, it's still possible that different releases or commits of that repo to build libroc and bindings.

> Do you have a backward compatibility promise that will absolutely guarantee that headers will not change?

Currently not. We *will* provide a compatibility promise starting from 1.0.0, according to semantic versioning. So, say, headers for 1.2.3 will be compatible with 1.2.4, 1.3.0, but not with 2.0.0.

Until 1.0.0. the library itself (libroc) as well as these bindings should be considered alpha.

> If not, maybe we can think about using submodules or something like that inthe future.

We can do it, but I don't think it will make any difference. We can periodically copy headers or we can periodically update submodule to a new commit. From the user point of view, a copy is a bit better than a submodule because the download size will be smaller.

Is there any other difference?

README.md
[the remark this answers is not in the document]
gavv: @xlab Hi,

> I'd not expect that bindings generation would work on any platform, it is supposed to work only on dev's machine. It's very optimistic to generate bindings dynamically upon installation.

Agree.

> I usually generate bindings that are common for each platform (using UNIX env on my macOS), and then add required differencies for win, pure linux, etc.

Could you elaborate? Does c-for-go generate some platform-dependent code? Can we generate everything on Linux and expect that go get will be able to build the generated code everywhere?

(Given that we will use traits when necessary)

roc.yml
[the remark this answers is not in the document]
gavv: @Asalle it seems we can remove the `/usr/lib/gcc/x86_64-redhat-linux/9/include` path if we will use `-ccincl` option:

    c-for-go -ccincl ./roc.yml

roc/types.go
gavv: Yep. I'd like to use a mixed approach. It would be convenient to export *some* generated stuff (e.g. configs) and unexport the rest and provide higher-level wrappers for it (https://github.com/roc-project/roc/pull/269#issuecomment-538951694).

README.md
```diff
@@ -6,3 +6,14 @@ Go bindings for Roc
 ```
 go get github.com/roc-project/roc-go/roc
 ```
+
+# Build
+1. Clone roc: `git clone git@github.com:roc-project/roc.git`
+2. Install SCons: `sudo dnf install scons`
```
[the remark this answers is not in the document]
Asalle: yes, as I said in the pr's body this readme is temporary and will be fixed before merging.

README.md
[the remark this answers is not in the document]
xlab: > Can we generate everything on Linux and expect that go get will be able to build the generated code everywhere?

Yes, this is the most likely scenario. In some cases there are some difference, however, if lib was tightly coupled with os internals, I think your lib is not the case.

Most difference in real apps is the set of LDFLAGS/CFLAGS. Also, sometimes it is easier to keep all traits in manually created file inside package. (e.g. https://github.com/vulkan-go/glfw/blob/v3.3/v3.3/glfw/build.go)

README.md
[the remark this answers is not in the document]
dshil: Yes, I saw your comment. See also https://github.com/roc-project/roc-go/pull/1#issuecomment-541526070

README.md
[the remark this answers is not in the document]
Asalle: I will remove the "traits:linux" thing from roc.yml, can we consider this thread resolved?

roc.yml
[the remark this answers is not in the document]
Asalle: works like charm @gavv. I added it to Makefile, can we consider the thread closed?

README.md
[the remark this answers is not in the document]
gavv: Yes.

roc.yml
Asalle: I removed them, so it's now on you @gavv that -lroc will be packaged with the same name on all platforms :)

roc.yml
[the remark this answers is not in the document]
gavv: Yep.

Asalle:
> Yes. But there would be two sources anyway: one copy of headers used when building libroc (by user or e.g. by deb-package maintainer) and another copy used when building bindings (by go get).

We (the go bindings) can also have a dependency on libroc-dev and demand installing it (which is now available only as sources, but will be probably packaged in the future). Usually go binding don't contain any c-headers, just the go files, right? c headers are needed only for generating the go bindings, so we can keep them for now and ditch them before merging to master, what do you think?
 
> We can do it, but I don't think it will make any difference. We can periodically copy headers or we can periodically update submodule to a new commit. From the user point of view, a copy is a bit better than a submodule because the download size will be smaller.
> 
> Is there any other difference?

With submodules you can at least be sure what revision you have, when copying such information is lost forever. But I think submodules will not be good as well, better let's not have cpp includes in the repo whatsoever, as I've written up there ^.

roc-wrapper.go
Asalle: done! check out my most recent commits!

roc/types.go
Asalle: yes, I think I can switch it on/off with c-for-go, but my attempts were unsuccessful so far...

Asalle:
> Replace LogHandler function with a Logger interface

I just saw that golang does not yet have a `logger` interface ([link](https://github.com/golang/go/issues/13182)), what is the benefit of doing it now since we don't know which exact interfaace will be accepted?

roc-wrapper.go
```diff
@@ -0,0 +1,83 @@
...
+func convertErr(err uint32, errstr string) error {
+	if err == 0 {
+		return nil
+	}
+	return fmt.Errorf("%s: %d", errstr, err)
+}
+
+// Init parses the `ip` and `port` and initializes the Address object
+func NewAddress(family Family, ip string, port int) (*Address, error) {
+	var a *Address
```
gavv: Shouldn't it be `a := new(Address)` ?

roc-wrapper.go
```diff
@@ -0,0 +1,83 @@
...
+	err := AddressInit(a, family, ip, port)
+	return a, convertErr(err, "Error when initializing an address")
+}
+
+// Family get Address family
+func (a *Address) Family() *Family {
```
gavv: `*Family` -> `Family`

roc-wrapper.go
```diff
@@ -0,0 +1,83 @@
...
+	return AddressFamily(a)
+}
+
+// Ip get Address ip
+func (a *Address) Ip() *Ip {
```
gavv: `Ip() *Ip` -> `IP() string`

(Go code style suggests avoiding camel-case for acronyms)

roc-wrapper.go
```diff
@@ -0,0 +1,83 @@
...
+	return AddressIp(a)
+}
+
+// Port get Address port
+func (a *Address) Port() *Port {
```
gavv: `*Port` -> `int`

roc-wrapper.go
```diff
@@ -0,0 +1,83 @@
...
+	return AddressPort(a)
+}
+
+func OpenContext(config *ContextConfig) *Context {
```
gavv: I suggest to make it `OpenContext(config *ContextConfig) (*Context, error)` for convenience (as well as other Open* funcs).

roc/const.go
```diff
@@ -0,0 +1,116 @@
...
+	LogNone  LogLevel = iota
+	LogError LogLevel = 1
+	LogInfo  LogLevel = 2
+	LogDebug LogLevel = 3
+	LogTrace LogLevel = 4
+)
+
+const (
+	// AddressSize as declared in roc/address.h:40
+	AddressSize = 64
```
gavv: Can we hide this constant? It makes sense only for the C API.

gavv:
> We (the go bindings) can also have a dependency on libroc-dev and demand installing it (which is now available only as sources, but will be probably packaged in the future). Usually go binding don't contain any c-headers, just the go files, right? c headers are needed only for generating the go bindings, so we can keep them for now and ditch them before merging to master, what do you think?

Sounds reasonable. This seems to be a better approach. So yes, let's remove headers from this repo and require them to be installed system-wide. I think we can do it in this PR.

gavv:
> I just saw that golang does not yet have a logger interface (link), what is the benefit of doing it now since we don't know which exact interfaace will be accepted?

Oh, I mean create our own interface compatible with the standard `log.Logger` struct, so the user could pass its instance to roc. Something very simple, like:

```
type Logger interface {
    Print(v ...interface{})
}
```

roc-wrapper.go
```diff
@@ -0,0 +1,83 @@
+package roc
```
gavv: Can we move this file into the `roc` directory?

AFAIK the common convention is to use the same name for the last component of the import path and the package itself. This repo is named `roc-go` so if the import path will be `github.com/roc-project/roc-go` and the imported package will be `roc`, it will be a bit confusing. For that reason I suggest to move the package into the `roc` directory, so that the import path will be `github.com/roc-project/roc-go/roc`.

What do you think?

roc/roc.go
```diff
@@ -1,3 +1,169 @@
...
+#include "../include/roc/receiver.h"
+#include "../include/roc/sender.h"
+#include <stdlib.h>
+#include "cgo_helpers.h"
+*/
+import "C"
+import "unsafe"
+
+// AddressInit function as declared in roc/address.h:86
+func AddressInit(address *Address, family Family, ip string, port int32) int32 {
```
gavv: Can we unexport all these functions (for which we're creating exported wrappers)?

roc-wrapper.go
Asalle: true

roc-wrapper.go
Asalle: done

roc-wrapper.go
Asalle: done

roc-wrapper.go
Asalle: done

roc-wrapper.go
Asalle: I don't think OpenContext returns an error. I think it just returns a pointer and you'll need to check if it's null.

roc/const.go
Asalle: yes, you pointed out some exported/unexported stuff before and I'm trying to make it work with c-for-go (basically enumerate what do we want exported and unexport everything else)

roc-wrapper.go
Asalle: sure, can do. I initially thought that c-for-go will remove the file, but it's not the case.

roc/roc.go
Asalle: yup, still working on it,

Asalle:
> Oh, I mean create our own interface compatible with the standard log.Logger struct, so the user could pass its instance to roc. Something very simple, like:

I tried to do it right now and I'm not sure if I follow: we have two funcs in roc that have something to do with logs: 
```
// LogSetLevel function as declared in roc/log.h:76
func LogSetLevel(level LogLevel) {
	clevel, _ := (C.roc_log_level)(level), cgoAllocsUnknown
	C.roc_log_set_level(clevel)
}

// LogSetHandler function as declared in roc/log.h:90
func LogSetHandler(handler LogHandler) {
	chandler, _ := handler.PassValue()
	C.roc_log_set_handler(chandler)
}
```
log.Logger does not have any log level, so I don't know what to do with the `LogSetLevel`. In the second, you want to accept log.Logger instead of LogHandler? And what do you want to do with it after?

Asalle:
> It seems that for every frame, it will perform at least one allocation on the C heap and a few allocations on the Go heap.
Writing and reading frames (roc_sender_write and roc_receiver_read) are hot paths. Ideally we should avoid allocations in these functions.
Both functions don't use the frame memory after the function returns. So although the frame memory is managed by Go GC, it should be safe to pass the raw pointer to that memory to these functions. So theoretically doing it without allocations is quite feasible.
If we can configure c-for-go to do this for us, it would be great. Otherwise, we can configure c-for-go to disable generating bindings for those two functions and write them by hand.

so do you want to allocate a big chunk of memory once in `SenderWrite` for example and then just pass the pointer to all the functions underneath and reuse the memory? like this, see below
```
func SenderWrite(sender []Sender, frame []Frame) int32 {
...
mem, err := C.calloc()
...
}

func unpackArgSFrame(x []Frame) (unpacked *C.roc_frame, allocs *cgoAllocMap, mem &[]byte) {
... 
mem[3] = 4
...
}
```

roc-wrapper.go
gavv: > I don't think OpenContext returns an error.

Exactly. I'm suggesting to add the error artificially, like:

```
func OpenContext(config *ContextConfig) (*Context, error) {
    ctx := ContextOpen(config)
    if ctx == nil {
        return nil, errors.New(...)
    }
    return ctx. nil
}
```

This may look a bit stupid from the implementation point of view, but this will give the API user a hint that the function can fail and errors should be checked. Otherwise it may seem like it can't fail since there is no error in the signature, which so typical in the Go world.

gavv:
> log.Logger does not have any log level, so I don't know what to do with the LogSetLevel. In the second, you want to accept log.Logger instead of LogHandler? And what do you want to do with it after?

I suggest to provide the following API to the user:

```
type Logger interface {
    Print(v ...interface{})
}

func LogSetLevel(level LogLevel)

func LogSetHandler(logger Logger)
```

So the only change is to switch from LogHandler, which is a function, to Logger, which is an interface compatible with standard logger.

Why? Just for convenience. The user will be able to pass the standard logger (or other compatible loggers) directly to our lib.

What to do with the provided Logger instance? Three things I guess:

* create a closure which signature is compatible with LogHandler and is implemented by calling logger.Print()

* store this closure in a private global variable so that GC will not collect it

* pass it to roc_log_set_handler()

gavv:
> so do you want to allocate a big chunk of memory once in SenderWrite for example and then just pass the pointer to all the functions underneath and reuse the memory? like this, see below

No. I was hoping that we can avoid allocations at all here.

So if the user gives us a []float slice, we get the pointer to the raw memory of the slice data (using unsafe I guess) and pass this pointer to roc_sender_write().

Theoretically this should work because roc_sender_write() does not store this pointer anywhere. It is guaranteed that this memory will not be accessed after the function returns, so this approach should not cause problems with GC (these problems are the reason why cgo uses C.calloc which I'd like to avoid).

BTW, I think this can be done in a separate PR. We could start with a simple implementation that works and do optimizations later. Up to you, of course.

gavv:
@Asalle Since we do not want to include roc headers (which are MPL-2) in this repo anymore, what do you think about changing the license to MIT? It doesn't really matter in case of the bindings, I think, but MIT is just more common for the Go world.

Asalle:
Also I was thinking: what do you think about putting the generated code into an `internal` package (called something like `roc-impl`) and and only bindings and types in actual `roc`package?

Asalle:
Ok, I think I got all your comments fixed, apart from the unnecessary allocation thing which I guess can be done in the nex pr. This means that I'll do the following:
1. Remove include/roc
2. Write tests for the bindings
3. Think about how to find includes with pkg-config
4. Fixup Readme
5. Change the license to MIT

roc/logger.go
```diff
@@ -0,0 +1,21 @@
...
+func LogSetLevel(level LogLevel) {
+	LogSetLevelImpl(level)
+}
+
+func LogSetHandler(l logger) {
+	handler = func(level LogLevel, component string, message string) {
+		l.Print(fmt.Sprintf("%s: %s", component, message))
+	}
+
+	LogSetHandlerImpl(handler)
```
gavv: I suggest to add level to the message as well.

Also, it seems we need to guard this with a mutex. libroc provides a thread-safe interface and LogSetHandlerImpl() itself is thread-safe. However LogSetHandler() is currently not thread-safe.

gavv:
> Remove include/roc
> Change the license to MIT

I made these two changes and force-pushed to master to keep the git history clean.

gavv:
> Write tests for the bindings
> Think about how to find includes with pkg-config
> Fixup Readme

BTW if you wish you could split this into several PRs as well.

We will also need to port API documentation from doxygen to godoc, but this can be done later too.

Asalle:
Ok, thanks! I'll just add a basic test here then and clean up stuff and make it ready for merge.

roc/logger.go
gavv: Also, let's rename these funcs to more natural "SetLogLevel" and "SetLogHandler".

gavv: Hm, it seems that LogSetHandlerImpl generated by c-for-go already stores the function into a global variable: https://github.com/roc-project/roc-go/blob/251b4476ed4152e41f4efac0448f304c9e84f615/roc/cgo_helpers.go#L682-L690

So we don't need to duplicate this code.

However, it also seems that the generated code is not thread-safe. We should learn whether c-for-go is capable of generating thread-safe code. If it's not, we should implement thread-safety in our wrappers. The goal is to make the entire API thread-safe, just like original C API.

Asalle:
btw, I just wrote a basic test and I saw that your error handling is much more intricate than code 0 is fine, others == error. I will need to rework that.

README.md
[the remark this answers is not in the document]
Asalle: fixed readme

roc/types.go
Asalle: done

README.md
```diff
@@ -1,8 +1,18 @@
 Go bindings for Roc
 -------------------
 
-*Work in progress!*
-
```
gavv: Let's keep the "Work in progress" note for now. We still need to add more tests, documentation, CI, etc.

roc.yml
```diff
@@ -0,0 +1,69 @@
...
+    - ../include/roc/config.h
+    - ../include/roc/context.h
+    - ../include/roc/frame.h
+    - ../include/roc/log.h
+    - ../include/roc/platform.h
+    - ../include/roc/receiver.h
+    - ../include/roc/sender.h
+  FlagGroups:
+    - {name: "CFLAGS", flags: [
+        "-I\"../include\"",
```
gavv: I guess we can remove this block? Since we're removing headers from the repo.

roc.yml
```diff
@@ -0,0 +1,69 @@
...
+  IncludePaths: ["include", "/usr/lib/gcc/x86_64-redhat-linux/9/include"]
+  SourcesPaths:
+    - include/roc/address.h
+    - include/roc/config.h
+    - include/roc/context.h
+    - include/roc/frame.h
+    - include/roc/log.h
+    - include/roc/platform.h
+    - include/roc/receiver.h
+    - include/roc/sender.h
```
gavv: As well as this one?

roc/types.go
gavv: Forgot to push?

gavv: Forgot to push?

Asalle:
Yes, wip line is needed.
Will we squash things later and force push in master?

I suggest we leave roc.yml for now and remove it when wrappers will be
fully tested. I generated code quite often, even yesterday (to unexpoet
things) I think it comes quite in handy

gavv:
> Will we squash things later and force push in master?

It would be great if you squash all commits in this PR before merging it. Usually we freely edit history in the PR while it's not merged but we're avoiding to edit history already merged to a public branch (though this happens sometimes with the develop branch; but this never happens with the master branch and it is even marked as "protected")

> I suggest we leave roc.yml for now and remove it when wrappers will be
fully tested. I generated code quite often, even yesterday (to unexpoet
things) I think it comes quite in handy

Hm, but why do you want to remove roc.yml? We should keep it (and also a makefile rule) so that we can always re-generate the code on fresh roc release.

Asalle:
Ok I misunderstood you. I thought you want to remove the roc.yml. I'm all
for keeping it

gavv:
Ah. By "this block" I meant the lines selected in the review comment:

> Comment on lines +6 to +17

roc/types.go
Asalle: yup.

README.md
Asalle: done

roc.yml
Asalle: I think this block will stay there anyway, we'll just need to pick up the headers from pkg-config somehow? Maybe let's leave it. These bindings will not work with the pkg-config right now whether you have this block (and the cgo comment in the generated files) or not. Maybe let's leave it for future reference?

roc.yml
Asalle: see my comment above. roc.yml will probably remain in the repo giving the possibility to generate the code. Why not leave it as is and quickly work on the next pr that includes pkg-config?

roc/types.go
Asalle: that means that we don't want to have two separate packages: one with purely generated code and one with bindings, right? if so, can we consider this convo resolved?

roc.yml
gavv: I see. If you want to keep it for future reference, I suggest to replace "../include/roc", which is now a nonexistent path, to "/usr/include/roc", which is where scons installs the headers currently (by default). It would be at least correct path :)

Asalle: will do :)

gavv:
@Asalle there are several open conversations currently. I assume that we want to resolve all of them before merge, but if you want to postpone some to future PRs, ping me when the request will be ready for review/merge.

Asalle:
Sure, I'm going through all the open loops. I will summarize the bigger
changes that will be better off in the next pr and mention you in my next
message.

roc/types.go
gavv: > that means that we don't want to have two separate packages

Yep, I'd prefer a single package.

> if so, can we consider this convo resolved?

Nope. This comment was about hiding `Frame` type, i.e. replacing it with `frame`, and replacing `Sender.Write(*Frame)` with type-safe wrappers like `Sender.WriteAudioBytes([]byte)` and `Sender.WriteAudioFloats([]float)`, as described in https://github.com/roc-project/roc/issues/248.

We can do it in future PRs, though.

Exporting something like this to the user:

```
type Frame struct {
	Samples       unsafe.Pointer
	SamplesSize   uint
}
```

would be too low-level and unfamiliar for a Go library (because of the unsafe.Pointer).

Asalle: ok, noted :+1:

roc/const.go
Asalle: done

roc/roc.go
Asalle: somehow, even if I tell c-for-go to unexport functions, it still has them exported :( Fixed manually

Asalle:
Do you want me to squash before or after review?
These are the things that I would like to do in the nex prs:
1. Hide Frame type from user (#2)
2. "Artificially" add errors along to returns to make code safer and more familiar to the user (#3)
3. Get headers from pkg-config (#4)
4. Add level to log messages, guard logHandler with the mutex (#5)
5. Rework error handling (#6)

Among which I would consider 3 and 5 the most urgent.

gavv:
> Do you want me to squash before or after review?

I think you can squash it. Chances are we will merge this. But I'll have a chance to look at it only next week.

Asalle:
Fine with me, I'll squash it now.

roc/types.go
gavv: #2

roc-wrapper.go
gavv: #3

roc/roc.go
gavv: I've created #7.

roc/logger.go
gavv: #5

gavv:
@Asalle thanks a lot for finishing this! Go bindings is a really wanted thing for me and now we're one step closer to it.

Do you have plans to continue working on this in future? I could add you to github team so that you can manage issues/PRs, etc.

Asalle:
Hey, I'm very happy you accepted this! I would love to continue work on this and be a part of your GitHub team :)

gavv:
Cool. Sent an invite.

Issue #10, opened 2019-11-13
Add more tests


gavv:
Ideally, we should cover all API with unit tests.

Asalle:
Also, we need integration tests for `roc_receiver_read` and `roc_receiver_write`

gavv:
Closing with respect to #19.

PR #15, opened 2020-01-18
Bindings for address.h/context.h/receiver.h

Asalle:


roc/roc.go
```diff
@@ -1,170 +1,59 @@
...
+		str = "\x00"
+	}
+	return str
+}
+
+// NewAddress parses the `ip`, `port` and `family` and initializes the Address object
+func NewAddress(family Family, ip string, port int) (*Address, error) {
+	a := new(Address)
+	var err error
+	a.raw = (*C.roc_address)(C.calloc(1, (C.size_t)(C.sizeof_roc_address)))
```
Asalle: there is a calloc, but I suppose this is what we would expect from a ctor?

gavv: No, we want to avoid calloc because it's not gc-collectable and requires manual free.

Instead, you can use a private byte array of size ROC_ADDRESS_SIZE. Cast it to unsafe pointer and pass to roc_address_* functions. Since they never use the pointer after returning, this is safe.

gavv: Example of this approach: https://github.com/hraban/opus/blob/v2/encoder.go#L148

gavv: Yeah, and one more thing. When we're passing Go memory to C and we can't be 100% sure that this memory is referenced somewhere in Go *after* the C function returns (e.g. we're holding a reference somewhere or using a variable after return), we should explicitly prolong its lifetime using runtime.KeepAlive(), to ensure that GC will not collect it until then.

Example:

```
func Foo(obj* Obj) {
    C.call_foo((C.obj*)obj)
    runtime.KeepAlive(obj) // without this call,
                  // GC can collect obj WHILE call_foo is running
}
```

OR:

```
func Foo(obj* Obj) int {
    C.call_foo((C.obj*)obj)
    return obj->x // no need for KeepAlive because we're using obj after return
}
```

If `obj` would be allocated in C memory, KeepAlive wasn't needed as well.

So we may need KeepAlive for `roc_address` and `*config` in some places. We don't need it for `roc_context`, `roc_sender`, and `roc_receiver`.

gavv:
LGTM.

Asalle:
Please review only the most recent commit.
Let me know if it's ok, I'll squash the last 3 commits and go on to write more bindings.

Asalle:
@gavv I need some invalid `context_config` examples with which I can produce an error from `context_open`, have you got any? 
```
/** Open a new context.
 *
 * Allocates and initializes a new context. May start some background threads.
 *
 * @b Parameters
 *  - @p config should point to an initialized config
 *
 * @b Returns
 *  - returns a new context if it was successfully created
 *  - returns NULL if the arguments are invalid
 *  - returns NULL if there are not enough resources
 */
ROC_API roc_context* roc_context_open(const roc_context_config* config);
``

**Comment by Asalle** [comment]
I don't know what are the correct numbers for `ReceiverConfig` fields, could you help me with that @gavv ?
Right now I put some random values, they don't seem to work:
```
			receiverConfig: ReceiverConfig{
				FrameSampleRate:         1,
				FrameChannels:           2,
				FrameEncoding:           1,
				AutomaticTiming:         5,
				ResamplerProfile:        4,
				TargetLatency:           2,
				MaxLatencyOverrun:       10,
				MaxLatencyUnderrun:      1,
				NoPlaybackTimeout:       67,
				BrokenPlaybackTimeout:   50,
				BreakageDetectionWindow: 5},
```

**Comment by Asalle** [comment]
Also, will I have to write some integration tests for `roc_receiver_read`? It seems that in order to check the functionality I will have to send some frames from outside and let `roc` consume it.

**Comment by gavv** [comment]
Hi, sorry for delay. Looking at this now.

File: roc/roc.go [hunk 2]
```diff
@@ -1,170 +1,142 @@
...
-
-// senderClose function as declared in roc/sender.h:192
-func senderClose(sender *Sender) int32 {
-	csender, _ := (*C.roc_sender)(unsafe.Pointer(sender)), cgoAllocsUnknown
-	__ret := C.roc_sender_close(csender)
-	__v := (int32)(__ret)
-	return __v
+	cfamily := (C.roc_family)(family)
+	ip = safeString(ip)
+	cip := C.CString(ip)
```
gavv:
```
// Go string to C string
// The C string is allocated in the C heap using malloc.
// It is the caller's responsibility to arrange for it to be
// freed, such as by calling C.free (be sure to include stdlib.h
// if C.free is needed).
func C.CString(string) *C.char
```

There is a leak here because CString performs malloc. Actually we don't need CString, we can just cast `ip` to `C.char*`.

roc/roc.go
```diff
@@ -1,170 +1,142 @@
...
+	}
+	return (int)(res), nil
+}
+
+// ============================================================================
+//                              context.h
+// ============================================================================
+
+func OpenContext(contextConfig *ContextConfig) (*Context, error) {
+	c := C.roc_context_open((*C.roc_context_config)(unsafe.Pointer(&contextConfig)))
```
gavv: And here, we can't just cast ContextConfig to roc_context_config. Although logically they have the same fields in the same order, it's not guaranteed that they have the same layout in memory. E.g. they may have different padding and alignment. Actually I'm not even sure that Go compiler guarantees to preserve the field order in memory.

So instead of a cast, we should allocate C.roc_context_config (better on stack) and do a field-by-field copy from Go struct to C struct.

roc/roc.go
```diff
@@ -1,170 +1,142 @@
...
+	if family == AfInvalid {
+		return family, ErrInvalidArguments
+	}
+	return family, nil
+}
+
+func (a *Address) IP() (string, error) {
+	const buflen = 255
+	sIP := make([]byte, buflen)
+	res := C.roc_address_ip(a.raw, (*C.char)(unsafe.Pointer(&sIP)), buflen)
```
gavv: We can allocate sIP on stack instead of the heap. I mean, use array instead of a slice.

gavv:
> Let me know if it's ok, I'll squash the last 3 commits and go on to write more bindings.

Github allows to see what changed since last review, so there is no need to keep new changes in separate commits *for review*. So feel free to organize commits whatever you want.

I'd suggest to use a single commit or maintain a few commits for different set of changes (new features, bug fixes, refactoring). After each change, you can just amend / rebase and force-push, it's OK.

> I need some invalid context_config examples with which I can produce an error from context_open, have you got any?

Look at the source code:

https://github.com/roc-project/roc/blob/master/src/lib/src/context.cpp#L23-L51

https://github.com/roc-project/roc/blob/master/src/lib/src/config.cpp#L17-L31

It seems that currently make_context_config() always returns true, so currently there is no such thing as invalid context config. This may change in future.

> I don't know what are the correct numbers for ReceiverConfig fields, could you help me with that @gavv ?

Take a look at the integration tests for the C API:

* https://github.com/roc-project/roc/tree/master/src/tests/roc_lib (non-real-time tests)

* https://github.com/roc-project/roc-tests (real-time tests)

And also at the C API examples:

* https://github.com/roc-project/roc/tree/master/src/lib/example

All of them provide examples of creating a context, a sender, and a receiver, and transferring samples from sender to receiver.

Also note that you can leave most of the fields set to zero, which means "use default value". Only a few fields don't have a default value and should be set explicitly. (It's documented in the API).

> Also, will I have to write some integration tests for roc_receiver_read? It seems that in order to check the functionality I will have to send some frames from outside and let roc consume it.

Exactly. You'll need to construct a sender and a receiver, write samples to sender, and read them from receiver.

Since Roc implement real-time streaming, the delivery is not 100% reliable. So if you run tests on a slow travis worker, for example, you will get drops like in UDP instead of delayed delivery like in TCP. At the receiver side drops will show as zero samples in the output stream. If a sample was not delivered in time, you'll read a zero instead of it.

For this reason, our non-real-time tests (which are intended to be run on travis, under valgrind, and so on), are not very strict. We don't expect that we will receive exactly all sent samples. We expect to receive only some of them and allow for drops and shifts. For more details, see https://github.com/roc-project/roc/issues/223 and https://github.com/roc-project/roc/issues/278

You'll need to reproduce this logic (maybe in some simple form) in your Go tests. We don't need to cover this in Go tests very carefully because we already covered it in C. However we need at least one small test that ensures that the receiver returns something adequate.

roc/errors.go
```diff
@@ -0,0 +1,37 @@
...
+	"errors"
+)
+
+var (
+	// ErrInvalidArguments indicates that one or more arguments passed to the function
+	// are invalid
+	ErrInvalidArguments = errors.New("One or more arguments are invalid")
+
+	// ErrInvalidApi should never happen and indicates that the API don't follow the declared contract
+	ErrInvalidApi = errors.New("nvalid return code from API")
```
gavv: `nvalid` -> `Invalid`

gavv:
> Also, will I have to write some integration tests for roc_receiver_read?

BTW, I think we can do it in a separate PR.

gavv:
Let me know when the PR is ready for further review.

roc/roc.go
gavv: Hm, no need for this, slice is okay. Go compiler is able to to allocate (small) slices on stack, even created using make: https://go.godbolt.org/z/2dg9Bf (chnage 255 to 255000 to see the difference)

Asalle:
> I need some invalid context_config

My mistake, I actually need some receiver_config example. Following your tip I took a look at `src/tests/roc_lib/test_sender_receiver.cpp`, at snippet here:

```
    void init_config(unsigned flags) {
        memset(&sender_conf, 0, sizeof(sender_conf));
        sender_conf.frame_sample_rate = SampleRate;
        sender_conf.frame_channels = ROC_CHANNEL_SET_STEREO;
        sender_conf.frame_encoding = ROC_FRAME_ENCODING_PCM_FLOAT;
        sender_conf.automatic_timing = 1;
        sender_conf.resampler_profile = ROC_RESAMPLER_DISABLE;
        sender_conf.packet_length =
            PacketSamples * 1000000000ul / (SampleRate * NumChans);
        if (flags & FlagFEC) {
            sender_conf.fec_code = ROC_FEC_RS8M;
            sender_conf.fec_block_source_packets = SourcePackets;
            sender_conf.fec_block_repair_packets = RepairPackets;
        } else {
            sender_conf.fec_code = ROC_FEC_DISABLE;
        }

        memset(&receiver_conf, 0, sizeof(receiver_conf));
        receiver_conf.frame_sample_rate = SampleRate;
        receiver_conf.frame_channels = ROC_CHANNEL_SET_STEREO;
        receiver_conf.frame_encoding = ROC_FRAME_ENCODING_PCM_FLOAT;
        receiver_conf.automatic_timing = 1;
        receiver_conf.resampler_profile = ROC_RESAMPLER_DISABLE;
        receiver_conf.target_latency = Latency * 1000000000ul / SampleRate;
        receiver_conf.no_playback_timeout = Timeout * 1000000000ul / SampleRate;
    }
```

And tried to replicate the config in my tests, which looks like this now:
```
			receiverConfig: ReceiverConfig{
				FrameSampleRate:         44100,
				FrameChannels:           ChannelSetStereo,
				FrameEncoding:           FrameEncodingPcmFloat,
				AutomaticTiming:         1,
				ResamplerProfile:        ResamplerDisable,
				TargetLatency:           0,
				MaxLatencyOverrun:       0,
				MaxLatencyUnderrun:      0,
				NoPlaybackTimeout:       0,
				BrokenPlaybackTimeout:   0,
				BreakageDetectionWindow: 0},
```

`target_latency` and `playback_timeout` are set to 0, unlike in your test, but it seems that I'm failing much earlier: this is the log I got from `roc` while executing the test:

```
14:51:36.303 [err] roc_lib: roc_config: invalid frame_channels
14:51:36.303 [err] roc_lib: roc_receiver_open: invalid arguments: bad config
14:51:36.303 [err] roc_lib: roc_receiver_close: invalid arguments: receiver is null
```
There is only one value that can be set as `FrameChannels` and it's `ChannelSetStereo`, I don't quite understand why I'm getting the error.

Asalle:
> BTW, I think we can do it in a separate PR.

Added it to #10

roc/errors.go
Asalle: fixed

roc/roc.go
Asalle: I don't think we can do it, I tried this:
`cip := (*C.char)(unsafe.Pointer(&ip))` and the test showed that address is nil (`NewAdress` failed)

what we can do is:
1) allocate and then free
2) make `[]C.char` of fixed size and copy byte-by-byte

roc/roc.go
Asalle: this is what `c-for-go` generated for the type `Context`:
```
type Context C.roc_context
```
shoud I rewrite the type as well?

roc/roc.go
Asalle: awesome!

Asalle:
> Let me know when the PR is ready for further review.

I wouldn't say it's ready but I need some input from you to move forward, see comments.

roc/roc.go
gavv: > cip := (*C.char)(unsafe.Pointer(&ip))

This doesn't work because `&ip` is the address of the *string object*, but not the address of the string data. (string object is a pointer to string data + its length, IIRC).

However, I just checked and we really can not cast string to `*C.char` because in Go you can't get an address of the string data (i.e. `&s[0]` is a compilation error).

But you can cast it to []byte and then cast (though it may cost an allocation):

```
$ cat 1.go 
package main

/*
#include <stdio.h>

void foo(const char* str) {
        printf("<%s>\n", str);
}
*/
import "C"
import (
        "runtime"
        "unsafe"
)

func Foo(s string) {
        b := []byte(s)
        C.foo((*C.char)(unsafe.Pointer(&b[0])))
        runtime.KeepAlive(b)
}

func main() {
        Foo("hello")
}
```

```
$ go run 1.go
<hello>
```

> allocate and then free
> make []C.char of fixed size and copy byte-by-byte

Agree, also good.

roc/roc.go
gavv: > type Context C.roc_context

This is OK, I guess. Both `C.roc_context` and `Context` are opaque types, so I see no problem here. We never dereference pointers to Context in Go code, just pass the pointer as is to the C code.

gavv:
> There is only one value that can be set as FrameChannels and it's ChannelSetStereo, I don't quite understand why I'm getting the error.

Your config looks good.

Before running the tests, you should fix the issues with passing objects from Go to C:

1. https://github.com/roc-project/roc-go/pull/15#discussion_r387201663

2. `(*C.roc_context)(unsafe.Pointer(&rocContext))` likely should be changed to `(*C.roc_context)(unsafe.Pointer(rocContext))` (rocContext is already a pointer, you pass *it* to C, but not its address)

Pay attention to every occurrence of unsafe.Pointer :)

It seems currently you're passing random memory to C.

roc/roc_test.go
```diff
@@ -0,0 +1,129 @@
+package roc
+
+import "testing"
+
+func fail(expected interface{}, got interface{}, t *testing.T) {
+	t.Errorf("Mismatch, expected: %v, got: %v", expected, got)
+	t.Fail()
+}
+
+func Test_roc_address_init(t *testing.T) {
```
gavv: I suggest to split roc_test.go into multiple files as well.

roc/roc.go
Asalle: done

Asalle:
> It seems currently you're passing random memory to C.

yup, you are right :joy_cat:

roc/roc_test.go
Asalle: done

gavv:
```
// WARNING: This file has automatically been generated on Sun, 15 Dec 2019 21:22:54 MSK.
// Code generated by https://git.io/c-for-go. DO NOT EDIT.
```

This comment, present in some files, is no more relevant.

roc/address.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func (a *Address) IP() (string, error) {
+	const buflen = 255
+	sIP := make([]byte, buflen)
+	res := C.roc_address_ip(a.raw, (*C.char)(unsafe.Pointer(&sIP)), buflen)
+	if res == nil {
+		return "", ErrInvalidArguments
+	}
+	return C.GoString(res), nil
```
gavv: Go compiler has no way to know that `res` points to `sIP` data, so here we should insert `runtime.KeepAlive` manually:

```
if res == nil {
	return "", ErrInvalidArguments
}
ret := C.GoString(res)
runtime.KeepAlive(sIP) // ret is actually a pointer to sIP data
return ret, nil
```

roc/address.go
```diff
@@ -0,0 +1,73 @@
...
+	a := new(Address)
+	a.mem = make([]byte, C.sizeof_roc_address)
+	a.raw = (*C.roc_address)(unsafe.Pointer(&a.mem))
+
+	cfamily := (C.roc_family)(family)
+	ip = safeString(ip)
+	cip := stringToCharArray(ip)
+	cport := (C.int)(port)
+	errCode := C.roc_address_init(a.raw, cfamily, (*C.char)(unsafe.Pointer(&cip[0])), cport)
+	runtime.KeepAlive(a.mem)
```
gavv: Hm, after more thought and googling, it seems that we don't need `runtime.KeepAlive` here.

Why:

* `unsafe.Pointer` references are traceable by GC
* so `*C.roc_address` should be traceable as well
* cgo automatically inserts `cgoUse()` call for every pointer argument, which is superior to `runtime.KeepAlive` (in addition, it forces the pointer to escape to heap, to prevent it moving during stack reallocation)

Thus, `a.mem` is guaranteed to be alive during `C.roc_address_init` call.

Why is this important? Because if it was wrong, we had to add `runtime.KeepAlive` to all other cgo calls, for every argument passed to C. And it seems that we don't have to, in most cases.

roc/receiver.go
```diff
@@ -0,0 +1,37 @@
...
+#include <stdlib.h>
+*/
+import "C"
+
+import (
+	"unsafe"
+)
+
+func OpenReceiver(rocContext *Context, receiverConfig *ReceiverConfig) (*Receiver, error) {
+	rocContextCPtr := (*C.roc_context)(unsafe.Pointer(rocContext))
```
gavv: Can we replace `(*C.roc_context)(unsafe.Pointer(rocContext))` with just `(*C.roc_context)(rocContext)`? And maybe get rid of the var at all.

roc/receiver.go
```diff
@@ -0,0 +1,37 @@
...
+
+	receiver := C.roc_receiver_open(rocContextCPtr, receiverConfigCPtr)
+	if receiver == nil {
+		return nil, ErrInvalidArguments
+	}
+	return (*Receiver)(receiver), nil
+}
+
+func (r *Receiver) Close() error {
+	errCode := C.roc_receiver_close((*C.roc_receiver)(unsafe.Pointer(r)))
```
gavv: Can we replace `(*C.roc_receiver)(unsafe.Pointer(r))` with just `(*C.roc_receiver)(r)`?

roc/address.go
```diff
@@ -0,0 +1,73 @@
+package roc
+
+/*
+#cgo LDFLAGS: -lroc
+#include <roc/receiver.h>
+#include <roc/address.h>
+#include <roc/sender.h>
+#include <roc/log.h>
+#include <stdlib.h>
```
gavv: Could you please cleanup C includes here and in other places? I don't think every go file needs all of them. Also, to we need `-lroc` in every file?

roc/address.go
```diff
@@ -0,0 +1,73 @@
...
+
+// NewAddress parses the `ip`, `port` and `family` and initializes the Address object
+func NewAddress(family Family, ip string, port int) (*Address, error) {
+	a := new(Address)
+	a.mem = make([]byte, C.sizeof_roc_address)
+	a.raw = (*C.roc_address)(unsafe.Pointer(&a.mem))
+
+	cfamily := (C.roc_family)(family)
+	ip = safeString(ip)
+	cip := stringToCharArray(ip)
```
gavv: Let's merge these two functions. Actually we can remove safeString() and just make the char array zero-terminated in stringToCharArray.

And then I'd rename `stringToCharArray` to something like `toCStr` and move to a separate file like `string.go` or `helpers.go`.

gavv:
I see that tests are passing despite of the memory issues I've mentioned above. Would be nice to cover these issues with tests, if possible.

roc/address.go
```diff
@@ -0,0 +1,73 @@
...
+		charArray[ind] = (C.char)(r)
+	}
+	return charArray
+}
+
+// NewAddress parses the `ip`, `port` and `family` and initializes the Address object
+func NewAddress(family Family, ip string, port int) (*Address, error) {
+	a := new(Address)
+	a.mem = make([]byte, C.sizeof_roc_address)
+	a.raw = (*C.roc_address)(unsafe.Pointer(&a.mem))
```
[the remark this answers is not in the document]
Asalle: I'm not sure how can I test it more, the test in `address_test.go` seeming tests everything there is for address

roc/address.go
```diff
@@ -0,0 +1,73 @@
...
+	if family == AfInvalid {
+		return family, ErrInvalidArguments
+	}
+	return family, nil
+}
+
+func (a *Address) IP() (string, error) {
+	const buflen = 255
+	sIP := make([]byte, buflen)
+	res := C.roc_address_ip(a.raw, (*C.char)(unsafe.Pointer(&sIP)), buflen)
```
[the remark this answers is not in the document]
Asalle: done

roc/receiver.go
```diff
@@ -0,0 +1,37 @@
...
+*/
+import "C"
+
+import (
+	"unsafe"
+)
+
+func OpenReceiver(rocContext *Context, receiverConfig *ReceiverConfig) (*Receiver, error) {
+	rocContextCPtr := (*C.roc_context)(unsafe.Pointer(rocContext))
+	receiverConfigCPtr := (*C.roc_receiver_config)(unsafe.Pointer(receiverConfig))
```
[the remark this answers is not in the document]
Asalle: done

roc/types.go
```diff
@@ -11,23 +11,20 @@ package roc
 #include "roc/sender.h"
 #include "roc/log.h"
 #include <stdlib.h>
-#include "cgo_helpers.h"
 */
 import "C"
 import "unsafe"
 
 // Address as declared in roc/address.h:59
 type Address struct {
```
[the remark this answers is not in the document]
Asalle: sounds reasonable, but maybe in the next pr?

roc/test-utils.go
```diff
@@ -0,0 +1,8 @@
+package roc
```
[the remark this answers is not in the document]
Asalle: done

roc/address.go
Asalle: hmm, how does the test work then? I would like to have something failing before fixing, because otherwise it's unclear if my code fixes the issue or not :thinking:

roc/address.go
Asalle: cool!

roc/receiver.go
Asalle: done

roc/receiver.go
Asalle: done

roc/address.go
Asalle: I need to investingate where the `-lroc` is needed, but otherwise includes are removed

roc/address.go
Asalle: done

roc/types.go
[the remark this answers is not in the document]
gavv: OK.

roc/address.go
gavv: It works because the probability that GC is called *between* C.roc_address_ip and C.GoString, *and* that it will collect sIP during this invocation, is very low. This makes such bugs very unpleasant because they likely will reproduce rarely and only under specific load conditions.

And it's very hard to cover runtime.KeepAlive stuff with tests. Maybe only with intensive stress testing. Adding a stress test for roc-go in future is a good idea I think.

So our main tool for now is very careful code review.

I hope that memory-related bugs (like using `&mem` instead of `&mem[0]`) are easier to catch with tests.

gavv: You can also try to run tests under race detector. We can even add it to the travis build.

Asalle: soo, here we shouldn't add keepAlive as well, following your arguments?
> unsafe.Pointer references are traceable by GC
    so *C.roc_address should be traceable as well
    cgo automatically inserts cgoUse() call for every pointer argument, which is superior to runtime.KeepAlive (in addition, it forces the pointer to escape to heap, to prevent it moving during stack reallocation)

roc/receiver.go
```diff
@@ -0,0 +1,38 @@
...
+
+func OpenReceiver(rocContext *Context, receiverConfig *ReceiverConfig) (*Receiver, error) {
+	receiverConfigC := C.struct_roc_receiver_config{
+		frame_sample_rate:         (C.uint)(receiverConfig.FrameSampleRate),
+		frame_channels:            (C.roc_channel_set)(receiverConfig.FrameChannels),
+		frame_encoding:            (C.roc_frame_encoding)(receiverConfig.FrameEncoding),
+		automatic_timing:          (C.uint)(receiverConfig.AutomaticTiming),
+		resampler_profile:         (C.roc_resampler_profile)(receiverConfig.ResamplerProfile),
+		target_latency:            (C.ulonglong)(receiverConfig.TargetLatency),
+		max_latency_overrun:       (C.ulonglong)(receiverConfig.MaxLatencyOverrun),
```
gavv, line 16: You forgot `MaxLatencyUnderrun`.

roc/context.go
```diff
@@ -0,0 +1,30 @@
+package roc
+
+/*
+#include <roc/receiver.h>
```
gavv, line 4: `roc/receiver.h` -> `roc/context.h`

Makefile
```diff
@@ -1,16 +1,13 @@
...
-	c-for-go -ccincl ./roc.yml
+.PHONY: check test
 
 check:
 	go build ./roc
 	golangci-lint run ./roc
 
 test:
-	go test ./roc
+	go test -race ./roc
```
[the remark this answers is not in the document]
gavv: BTW, better do it in separate travis steps.

gavv:
Looks almost good.

roc/address.go
Asalle: somehow this conversation gets hidden by github, that's why you probably aswered this question twice :thinking: 
I see your point :+1:

gavv: Yeah, github duplicates a review comment if it's an answer to a hidden discussion. Unfortunately github still doesn't work very well with long discussions and review threads.

roc/receiver.go
Asalle: fixed

roc/context.go
Asalle: fixed

gavv:
Here we go!

PR #18, opened 2020-06-02
Fix and improve travis build

gavv:
- freeze linter version
- add latest Go (1.x)
- add macOS


gavv:
@Asalle could you please take a look?

Issue #16, opened 2020-04-05
Add go.mod


gavv:
Add go.mode and use semantic versioning. Add a corresponding note to README.

PR #21, opened 2020-06-06
go.mod: add newly created go.mod

Asalle:
#16

### DIFF
--- a/roc/go.mod
+++ b/roc/go.mod
@@ -1,0 +1,3 @@
+module github.com/roc-streaming/roc-go/roc
+
+go 1.12
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,11 @@ go:
   - 1.x
 
 env:
-  - GOTESTFLAGS=""
-  - GOTESTFLAGS="-race"
+  global:
+    - GO111MODULE=on
+  jobs:
+    - GOTESTFLAGS=""
+    - GOTESTFLAGS="-race"
 
 jobs:
   exclude:
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,6 @@
+GO111MODULE := on
+export GO111MODULE
+
 all: check test race
 
 .PHONY: check


go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/roc-project/roc-go
```
gavv, line 1: Shouldn't we place this file into `roc` and rename the module to `github.com/roc-project/roc-go/roc`?

go.mod
```diff
@@ -0,0 +1,3 @@
...
+
+go 1.14
```
gavv, line 3: We actually support at least go 1.12 and it's even checked in our travis build.

go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/roc-project/roc-go
+
```
gavv, line 2: Shouldn't we also add GO111MODULE=on to `env` section in travis?

go.mod
Asalle: I'm not sure why do want to do this? do we plan to have more folders in `github.com/roc-project/roc-go/` with their own `go.mod`s? or is it just for the sake of changing the module name to `roc`?

go.mod
Asalle: alright

go.mod
Asalle: alright

go.mod
Asalle: I'm wondering how did the ci pass :thinking:

gavv: Hm, mys suggestion was not quite correct. By just adding a line to `env` we tell travis to build all combinations of these three exports:

```
  - GOTESTFLAGS=""
  - GOTESTFLAGS="-race"
  - GO111MODULE=on
```

image

We should use `global`: https://docs.travis-ci.com/user/environment-variables/#global-variables

go.mod
gavv: My bad, my understanding of go modules was wrong. I though that one module == one package, but that's not true. Let's keep it as is.

Asalle: alright!

go.mod
Asalle: fixed

gavv: The build matrix is broken now for some reason.

Was: https://travis-ci.org/github/roc-project/roc-go/builds/699219954

image

Now: https://travis-ci.org/github/roc-project/roc-go/builds/699428591

image

gavv: Maybe try to remove `- ` from `- global:` and `- jobs`.

gavv: https://config.travis-ci.com/explore

Asalle: I'll take a look into it

Asalle: yes, don't know what I was thinking

gavv:
Forgot one more detail, please also add GO111MODULE=on to Makefile, like here: https://github.com/gavv/webrtc-cli/blob/master/Makefile#L1-L2

gavv:
Sorry, one more concern.

It seems that moving go.mod to roc/ still makes sense. When we'll decide to tag v2 version (or v3, or v4, ...), we'll have to append it to go module name in go.mod, like:

```
module github.com/gavv/roc-go/v2
```

If go.mod is in project root, the user import will look like:

```
import "github.com/gavv/roc-go/v2/roc"
```

If, however, go.mod is in roc/ sub-directory, the user import would be:

```
import "github.com/gavv/roc-go/roc/v2"
```

The second form is more familiar, in my opinion. For most go packages, you just append the version to the import path. The first form is usually used only for multi-package repositories, however our repo is single-package.

Thus, I'd prefer the second form for roc-go. So maybe move go.mod to roc/ directory after all, what do you think?

Asalle:
@gavv then, maybe it makes sense to remove the folder and just put all the files in the project root? like here https://github.com/alicebob/miniredis

gavv:
> then, maybe it makes sense to remove the folder and just put all the files in the project root?

Technically possible, but in this case another inconsistency appears: the repo and the import path will be ".../roc-go", but the package name will be "roc". Which is allowed but considered bad practice. (Probably we already discussed it... Don't remember.)

gavv:
I've renamed github org (see mailing list), so we also need to update go.mod accordingly.

Asalle:
> I've renamed github org (see mailing list), so we also need to update go.mod accordingly.

cool, will change

alright, then let's make it `roc` with go.mod inside the folder

roc/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/roc-streaming/roc-go
```
gavv, line 1: Well it should be `github.com/roc-streaming/roc-go/roc` then? :)

Asalle: that's right

PR #22, opened 2020-06-07
Implement logger

### LINKED ISSUE #5: Implement Logger
Hm, it seems that LogSetHandlerImpl generated by c-for-go already stores the function into a global variable:

roc-go/roc/cgo_helpers.go
Lines 682 to 690 in 251b447
```
 func (x LogHandler) PassValue() (ref C.roc_log_handler, allocs *cgoAllocMap) { 
 	if x == nil { 
 		return nil, nil 
 	} 
 	if logHandler9E4BBC5EFunc == nil { 
 		logHandler9E4BBC5EFunc = x 
 	} 
 	return (C.roc_log_handler)(C.roc_log_handler_9e4bbc5e), nil 
 } 
```
So we don't need to duplicate this code.

However, it also seems that the generated code is not thread-safe. We should learn whether c-for-go is capable of generating thread-safe code. If it's not, we should implement thread-safety in our wrappers. The goal is to make the entire API thread-safe, just like original C API.

gavv:
Resolves #5.

- add SetLogLevel(), SetLoggerFunc(), and SetLogger()
- SetLogger() uses Logger interface compatible with log.Logger
- by default use standard logger instead of plain stderr
- by default disable logging in tests

A note on tests: roc logging is currently synchronous, but likely will
become asynchronous in future; logger tests are designed to work well
in both cases.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ all: check test race
 .PHONY: check
 check:
 	go build ./roc
+	go test ./roc -run xxx
 	golangci-lint run ./roc
 
 .PHONY: test
--- a/roc/log_test.go
+++ b/roc/log_test.go
@@ -1,0 +1,105 @@
+package roc
+
+import (
+	"io/ioutil"
+	"log"
+	"testing"
+	"time"
+)
+
+const defaultLogLevel LogLevel = LogError
+
+type testWriter struct {
+	ch chan string
+}
+
+func makeTestWriter() testWriter {
+	return testWriter{
+		// capacity 1 is needed to ensure that at least one
+		// message can be written without blocking
+		ch: make(chan string, 1),
+	}
+}
+
+func (tw testWriter) Write(buf []byte) (int, error) {
+	select {
+	case tw.ch <- string(buf):
+	default:
+		// drop message instead of blocking if the channel is full
+	}
+	return len(buf), nil
+}
+
+func (tw testWriter) wait() string {
+	const waitTimeout = time.Minute
+
+	select {
+	case s := <-tw.ch:
+		return s
+	case <-time.After(waitTimeout):
+		return ""
+	}
+}
+
+func Test_roc_log_default(t *testing.T) {
+	setupFuncs := []func(){
+		func() {},
+		func() { SetLoggerFunc(nil) },
+		func() { SetLogger(nil) },
+	}
+
+	SetLogLevel(LogDebug)
+	defer SetLogLevel(defaultLogLevel)
+
+	for n, setupFn := range setupFuncs {
+		tw := makeTestWriter()
+
+		log.SetOutput(&tw)
+		defer log.SetOutput(ioutil.Discard)
+
+		setupFn()
+
+		ctx, _ := OpenContext(&ContextConfig{})
+		ctx.Close()
+
+		if tw.wait() == "" {
+			t.Fatalf("test %v: expected logs, didn't get them before timeout", n)
+		}
+	}
+}
+
+func Test_roc_log_func(t *testing.T) {
+	SetLogLevel(LogDebug)
+	defer SetLogLevel(defaultLogLevel)
+
+	tw := makeTestWriter()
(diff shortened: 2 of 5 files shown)


gavv:
https://travis-ci.org/github/roc-project/roc-go/builds/695615095

Makefile
```diff
@@ -3,6 +3,7 @@ all: check test race
 .PHONY: check
 check:
 	go build ./roc
+	go test ./roc -run xxx
```
Asalle, line 6: just curious: we don't have a test that's called `xxx` - does this code checks (compiles) test files?

roc/log_test.go
```diff
@@ -0,0 +1,100 @@
...
+	default:
+	}
+	return len(buf), nil
+}
+
+func (tw testWriter) wait() string {
+	select {
+	case s := <-tw.ch:
+		return s
+	case <-time.After(time.Minute):
```
Asalle, line 34: maybe it's nice to have `time.Minute` as a named constant? makes things even clearer

roc/log_test.go
```diff
@@ -0,0 +1,100 @@
...
+		log.SetOutput(&tw)
+		defer log.SetOutput(ioutil.Discard)
+
+		setupFn()
+
+		ctx, _ := OpenContext(&ContextConfig{})
+		ctx.Close()
+
+		if tw.wait() == "" {
+			t.Fatalf("test %v failed", n)
```
Asalle, line 61: maybe it makes sense to say why exactly it failed: "Error: expected logs, didn't get them before timeout" or soemthing more articulate

roc/log_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+const defaultLogLevel LogLevel = LogError
+
+type testWriter struct {
+	ch chan string
+}
+
+func makeTestWriter() testWriter {
+	return testWriter{
+		ch: make(chan string, 1),
```
Asalle, line 18: I think 1 is the default value, so you don't have to specify it here. But we can keep it like this too, this is not an error

roc/log_test.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+}
+
+func (tw testWriter) Write(buf []byte) (int, error) {
+	select {
+	case tw.ch <- string(buf):
```
Asalle, line 24: why do you select here? isn't it the same as to do `tw.ch <- string(buf)`? Or you're afraid that it's going to block as nobody is reading it form the other side?

Makefile
gavv: Yep. `-run xxx` is just a convention to say "compile and run nothing", though maybe not a widely used one. I added this check because otherwise if test are not compiling, the linter in `make check` will produce cryptic errors.

roc/log_test.go
gavv: Nope, default value is 0: https://play.golang.org/p/YTtzluwniT6

roc/log_test.go
gavv: Done.

roc/log_test.go
gavv: Agree, added message here and in other tests.

roc/log_test.go
gavv: Also added a comment.

roc/log_test.go
gavv: To ensure that Write() wont block. Otherwise, if Roc will print logs syncrhonously, OpenContext() will hang when trying to print more than cap(tw.ch) messages, i.e more than one message. We either need a channel with big capacity or we can drop messages if the channel is full. I followed the second approach here.

I've added a comment.

gavv:
https://travis-ci.org/github/roc-project/roc-go/builds/695634123

roc/log_test.go
Asalle: I see, makes sense!

roc/log_test.go
Asalle: all clear now

PR #24, opened 2020-06-22
Impl bindings for receiver, sender

Asalle:
We don't want to expose the low-level `frame` structure, so both receiver and
sender use functions like `ReadBytes` and `WriteBytes` that operate on []byte.

I also added a simple "integration test" that writes something with sender,
reads it with receiver and checks that what we received is not an array of 0s.
Similar smoke-test can be found
[here](https://github.com/roc-project/roc-tests/blob/master/tests/test_service_quality.cpp).

The test does not pass somehow, maybe somebody can give me a hint?

### DIFF
--- a/roc/receiver.go
+++ b/roc/receiver.go
@@ -1,7 +1,12 @@
 package roc
 
 /*
-#include <roc/receiver.h>
+ #include <roc/receiver.h>
+ #include <roc/config.h>
+int rocGoReceiverReadFloats(roc_receiver* receiver, float* samples, unsigned long samples_size) {
+    roc_frame frame = {(void*)samples, samples_size*sizeof(float)};
+    return roc_receiver_read(receiver, &frame);
+}
 */
 import "C"
 
@@ -32,6 +37,38 @@ func OpenReceiver(rocContext *Context, receiverConfig *ReceiverConfig) (*Receive
 		return nil, ErrInvalidArgs
 	}
 	return (*Receiver)(receiver), nil
+}
+
+func (r *Receiver) Bind(portType PortType, proto Protocol, a *Address) error {
+	errCode := C.roc_receiver_bind(
+		(*C.roc_receiver)(r),
+		(C.roc_port_type)(portType),
+		(C.roc_protocol)(proto),
+		a.raw)
+	if errCode == 0 {
+		return nil
+	}
+	if errCode < 0 {
+		return ErrInvalidArgs
+	}
+	panic(fmt.Sprintf(
+		"unexpected return code %d from roc_receiver_bind()", errCode))
+}
+
+func (r *Receiver) ReadFloats(frame []float32) error {
+	if frame == nil {
+		return ErrInvalidArgs
+	}
+	errCode := C.rocGoReceiverReadFloats((*C.roc_receiver)(r), (*C.float)(&frame[0]), (C.ulong)(len(frame)))
+	if errCode == 0 {
+		return nil
+	}
+
+	if errCode < 0 {
+		return ErrInvalidArgs
+	}
+	panic(fmt.Sprintf(
+		"unexpected return code %d from roc_receiver_read()", errCode))
 }
 
 func (r *Receiver) Close() error {
--- a/roc/sender-receiver_test.go
+++ b/roc/sender-receiver_test.go
@@ -1,0 +1,169 @@
+package roc
+
+import (
+	"sync"
+	"testing"
+)
+
+type testEnv struct {
+	Sender   *Sender
+	Receiver *Receiver
+	Context  *Context
+}
+
+func makeReceiverConfig() *ReceiverConfig {
+	return &ReceiverConfig{
+		FrameSampleRate:  44100,
+		FrameChannels:    ChannelSetStereo,
+		FrameEncoding:    FrameEncodingPcmFloat,
+		AutomaticTiming:  1,
+		ResamplerProfile: ResamplerDisable,
+	}
+}
+
+func makeSenderConfig() *SenderConfig {
+	return &SenderConfig{
+		FrameSampleRate:  44100,
+		FrameChannels:    ChannelSetStereo,
+		FrameEncoding:    FrameEncodingPcmFloat,
+		AutomaticTiming:  1,
+		ResamplerProfile: ResamplerDisable,
+		FecCode:          FecRs8m,
+	}
+}
+
+func newTestEnv(t *testing.T) *testEnv {
+	var (
+		err error
+		e   testEnv
+	)
+	// create context
+	e.Context, err = OpenContext(&ContextConfig{})
+	if e.Context == nil || err != nil {
+		t.Errorf("Cannot create context: %v, error: %v", e.Context, err)
+	}
+
+	// create receiver
+	e.Receiver, err = OpenReceiver(e.Context, makeReceiverConfig())
+	if e.Receiver == nil || err != nil {
+		t.Errorf("Cannot create receiver: %v, error: %v", e.Receiver, err)
+	}
+
+	// bind receiver to two ports
+	// 1)
+	sourceAddr, err := NewAddress(AfAuto, "127.0.0.1", 0)
+	if sourceAddr == nil || err != nil {
+		t.Errorf("Cannot create source address object, address object: %v, error: %v",
+			sourceAddr, err)
+	}
+	err = e.Receiver.Bind(PortAudioSource, ProtoRtpRs8mSource, sourceAddr)
+	if err != nil {
+		t.Errorf("Cannot bind receiver: %v", err)
+	}
+
+	// 2)
+	repairAddr, err := NewAddress(AfAuto, "127.0.0.1", 0)
+	if repairAddr == nil || err != nil {
+		t.Errorf("Cannot create repair address object, address object: %v, error: %v",
+			repairAddr, err)
+	}
+	err = e.Receiver.Bind(PortAudioRepair, ProtoRs8mRepair, repairAddr)
+	if err != nil {
+		t.Errorf("Cannot bind receiver: %v", err)
+	}
+
+	// create sender
+	e.Sender, err = OpenSender(e.Context, makeSenderConfig())
+	if e.Sender == nil || err != nil {
+		t.Errorf("Cannot create sender, sender: %v, error: %v", e.Sender, err)
+	}
+
+	// bind sender to a port
+	senderAddr, err := NewAddress(AfAuto, "127.0.0.1", 0)
+	if senderAddr == nil || err != nil {
+		t.Errorf("Cannot create sender address object, address object: %v, error: %v",
+			senderAddr, err)
+	}
+	err = e.Sender.Bind(senderAddr)
+	if err != nil {
+		t.Errorf("Cannot bind sender: %v", err)
+	}
+
+	// connect sender to receiver ports
+	// 1)
+	err = e.Sender.Connect(PortAudioSource, ProtoRtpRs8mSource, sourceAddr)
+	if err != nil {
+		t.Errorf("Cannot connect sender to receiver: %v", err)
+	}
+	// 2)
+	err = e.Sender.Connect(PortAudioRepair, ProtoRs8mRepair, repairAddr)
+	if err != nil {
+		t.Errorf("Cannot connect sender to receiver: %v", err)
+	}
+	return &e
+}
+
+func (e *testEnv) close(t *testing.T) {
+	err := e.Receiver.Close()
+	if err != nil {
+		t.Fail()
+	}
+	err = e.Sender.Close()
+	if err != nil {
+		t.Fail()
+	}
+	err = e.Context.Close() // remove after finalizers are done
+	if err != nil {
+		t.Fail()
+	}
+}
+
+func Test_roc_sender_write_receiver_read(t *testing.T) {
+	e := newTestEnv(t)
+	defer e.close(t)
+
+	samplesCnt := 2
+	samples := make([]float32, samplesCnt)
+	for i := 0; i < samplesCnt; i++ {
+		samples[i] = float32(i + 1)
+	}
+
+	endChan := make(chan struct{})
+	var wait sync.WaitGroup
+	wait.Add(1)
+	go func() {
+		for {
+			err := e.Sender.WriteFloats(samples)
+			if err != nil {
+				t.Fail()
+			}
+			select {
+			case <-endChan:
+				wait.Done()
+				return
+			default:
+			}
+		}
+	}()
+
+	for {
+		recFloats := make([]float32, samplesCnt)
+		err := e.Receiver.ReadFloats(recFloats)
+		if err != nil {
+			t.Fail()
+		}
+
+		nonZeroSamplesCnt := 0
+		for i := 0; i < samplesCnt; i++ {
+			if recFloats[i] != 0 {
+				nonZeroSamplesCnt++
+			}
+		}
+		if nonZeroSamplesCnt == samplesCnt {
+			break
+		}
+	}
+
+	endChan <- struct{}{}
+	wait.Wait()
+}
--- a/roc/sender.go
+++ b/roc/sender.go
@@ -2,8 +2,95 @@ package roc
 
 /*
 #include <roc/sender.h>
+int rocGoSenderWriteFloats(roc_sender* sender, float* samples, unsigned long samples_size) {
+    roc_frame frame = {(void*)samples, samples_size*sizeof(float)};
+    return roc_sender_write(sender, &frame);
+ }
 */
 import "C"
 
+import (
+	"fmt"
+)
+
 // Sender as declared in roc/sender.h:96
 type Sender C.roc_sender
+
+func OpenSender(rocContext *Context, senderConfig *SenderConfig) (*Sender, error) {
+	senderConfigC := C.struct_roc_sender_config{
+		frame_sample_rate:        (C.uint)(senderConfig.FrameSampleRate),
+		frame_channels:           (C.roc_channel_set)(senderConfig.FrameChannels),
+		frame_encoding:           (C.roc_frame_encoding)(senderConfig.FrameEncoding),
+		packet_sample_rate:       (C.uint)(senderConfig.PacketSampleRate),
+		packet_channels:          (C.roc_channel_set)(senderConfig.PacketChannels),
+		packet_encoding:          (C.roc_packet_encoding)(senderConfig.PacketEncoding),
+		packet_length:            (C.ulonglong)(senderConfig.PacketLength),
+		packet_interleaving:      (C.uint)(senderConfig.PacketInterleaving),
+		automatic_timing:         (C.uint)(senderConfig.AutomaticTiming),
+		resampler_profile:        (C.roc_resampler_profile)(senderConfig.ResamplerProfile),
+		fec_code:                 (C.roc_fec_code)(senderConfig.FecCode),
+		fec_block_source_packets: (C.uint)(senderConfig.FecBlockSourcePackets),
+		fec_block_repair_packets: (C.uint)(senderConfig.FecBlockRepairPackets),
+	}
+	sender := C.roc_sender_open((*C.roc_context)(rocContext), &senderConfigC)
+	if sender == nil {
+		return nil, ErrInvalidArgs
+	}
+	return (*Sender)(sender), nil
+}
+
+func (s *Sender) Bind(a *Address) error {
+	errCode := C.roc_sender_bind((*C.roc_sender)(s), a.raw)
+	if errCode == 0 {
+		return nil
+	}
+	if errCode < 0 {
+		return ErrInvalidArgs
+	}
+	panic(fmt.Sprintf(
+		"unexpected return code %d from roc_address_init()", errCode))
+}
+
+func (s *Sender) Connect(portType PortType, proto Protocol, address *Address) error {
+	errCode := C.roc_sender_connect(
+		(*C.roc_sender)(s),
+		(C.roc_port_type)(portType),
+		(C.roc_protocol)(proto),
+		address.raw,
+	)
+	if errCode == 0 {
+		return nil
+	}
+	if errCode < 0 {
+		return ErrInvalidArgs
+	}
+	panic(fmt.Sprintf(
+		"unexpected return code %d from roc_address_init()", errCode))
+}
+
+func (s *Sender) WriteFloats(frame []float32) error {
+	if frame == nil {
+		return ErrInvalidArgs
+	}
+	errCode := C.rocGoSenderWriteFloats((*C.roc_sender)(s), (*C.float)(&frame[0]), (C.ulong)(len(frame)))
+	if errCode == 0 {
+		return nil
+	}
+	if errCode < 0 {
+		return ErrInvalidArgs
+	}
+	panic(fmt.Sprintf(
+		"unexpected return code %d from roc_address_init()", errCode))
+}
+
+func (s *Sender) Close() error {
+	errCode := C.roc_sender_close((*C.roc_sender)(s))
+	if errCode == 0 {
+		return nil
+	}
+	if errCode < 0 {
+		return ErrInvalidArgs
+	}
+	panic(fmt.Sprintf(
+		"unexpected return code %d from roc_address_init()", errCode))
+}


roc/sender-receiver_test.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	// connect sender to receiver ports
+	// 1)
+	err = sender.Connect(PortAudioSource, ProtoRtpRs8mSource, sourceAddr)
+	if err != nil {
+		t.Errorf("Cannot connect sender to receiver: %v", err)
+	}
+	// 2)
+	err = sender.Connect(PortAudioSource, ProtoRtpRs8mSource, repairAddr)
```
gavv, line 82: It should be:

```
sender.Connect(PortAudioRepair, ProtoRs8mRepair, repairAddr)
```

roc/receiver.go
```diff
@@ -30,6 +35,63 @@ func OpenReceiver(rocContext *Context, receiverConfig *ReceiverConfig) (*Receive
...
+	if errCode == 0 {
+		return nil
+	}
+	if errCode < 0 {
+		return ErrInvalidArguments // roc_receiver_bind returns negative in 3 cases: invalid args, cannot bind address and not enough resources, maybe we should consider renaming this error to a more generic name?
+	}
+	return ErrInvalidApi
+}
+
+func (r *Receiver) ReadAudioBytes(bytes []byte, samplesSize uint) error {
```
gavv, line 53: I think right now reading or writing bytes doesn't make much sense. The only frame encoding supported so far is FrameEncodingPcmFloat, i.e. floats. So I suggest to remove bytes versions of read and write for now.

roc/receiver.go
```diff
@@ -30,6 +35,63 @@ func OpenReceiver(rocContext *Context, receiverConfig *ReceiverConfig) (*Receive
...
+		return nil
+	}
+
+	if errCode < 0 {
+		return ErrInvalidArguments // roc_receiver_bind returns negative in 3 cases: invalid args, cannot bind address and not enough resources, maybe we should consider renaming this error to a more generic name?
+	}
+	return ErrInvalidApi
+}
+
+func (r *Receiver) ReadAudioFloats(floats []float32, samplesSize uint) error {
```
gavv, line 74: 1. I suggest to rename it to just `ReadFloats` for clarity.
2. `samplesSize` parameter is redundant - we can just use `len(floats)` (like io.Reader does)

roc/sender.go
```diff
@@ -1,9 +1,106 @@
...
+	if errCode == 0 {
+		return nil
+	}
+	if errCode < 0 {
+		return ErrInvalidArguments // roc_receiver_bind returns negative in 3 cases: invalid args, cannot bind address and not enough resources, maybe we should consider renaming this error to a more generic name?
+	}
+	return ErrInvalidApi
+}
+
+func (s *Sender) WriteFloats(frame []float32, samplesSize uint) error {
```
gavv, line 81: samplesSize is not needed here too

roc/sender-receiver_test.go
```diff
@@ -0,0 +1,127 @@
...
+		b[i] = byte(i)
+	}
+
+	err := sender.WriteBytes(b, 1)
+	if err != nil {
+		t.Fail()
+	}
+
+	recBytes := make([]byte, bytesCnt)
+	err = receiver.ReadAudioBytes(recBytes, 1)
```
gavv, line 104: This piece of code has several problems:

1. Writing raw bytes is dangerous. In makeSenderConfig we set FrameEncoding to FrameEncodingPcmFloat (the only supported encoding currently), which mean that we should write floats, not bytes. We also set FrameChannels to ChannelSetStereo (the only supported channel set currently), which means that we should write samples in interleaved format, i.e. pairs of L and R samples. So the minimum unit we can pass to roc_sender_write is two floats. Floats are 32 bit, so the minimum unit is 8 bytes. And the frame size should be multiple of 8 bytes too. So our write and read calls are incorrect. I suggest to get rid of byte functions at all and use WriteFloats/ReadFloats here.

2. The stream read from receiver is not guaranteed to be identical to the stream written to sender. First, received stream will have some leading zeros, not present in sent stream, because of the jitter buffer latency. Second, received stream may contain zeros instead of some samples if corresponding packet was lost and receiver wasn't able to recover it (or FEC is disabled). Third, received stream may contain shifts, if the session was restarted in the middle of a stream, which happens e.g. on high loss rate. This is not a frequent event in real use cases, however it happens from time to time, including travis workers which are sometimes very loaded and thus slow and thus packet loss increases. Forth, if resampling is enabled, sample rate conversion may change actual sample values a bit. The difference is not audible, but the samples are not equal numbers any more even if there is no packet loss.

3. Sender and receiver streams are both real-time and they don't block like in TCP. Receiver read does not wait until something is received. If some samples were not received in time, they're just replaced with zeros. If no samples were received, or there are no senders at all, receiver just produces all zeros. So if you call receiver read in a loop and write a single non-zero sample to sender (say, from another thread), receiver will produce a stream of zeros with a single non-zero sample *somewhere in the stream*. It's not guaranteed to be the first sample you'll get from receiver. And it's not guaranteed to be present at all because we use UDP, which is unreliable. See also my comment here.

gavv: The idea of libroc is to provide a simple TCP-like API and hide the complexity of real-time streaming behind it.

The user can just write samples, read samples, and libroc will try to fulfill two contradicting requirements: 1) maintain good service quality 2) maintain fixed latency. In other words, it will try to "sound good". But that's all. We don't (and can't) provide stronger guarantees like in TCP. This is the cost we pay for real-time.

So the API is TCP-like but the behavior is not. You can expect that it will sound good, but you can't make much assumptions of the exact content on the received stream.

BTW, in future we likely will add more transports, including reliable ones like HTTP. But so far this is not our major use case.

gavv: To make this test reliable, we need to send samples in a loop and read samples in a loop, and wait until we receive at least N non-zero samples. The same technique was implemented in https://github.com/roc-project/roc/pull/283. Please take a look at that issue and discussions in the PR, in particular this one: https://github.com/roc-project/roc/pull/283#discussion_r341954058

gavv: Yes, and regarding the test here: https://github.com/roc-project/roc-tests/blob/master/tests/test_service_quality.cpp

roc-tests repo is intended to collect *real-time* tests. They wont be 100% reliable. Their result wont be just boolean like "passed" or "failed", but instead something like "the loss ration was below 5% this time". And they will be run on real hardware on unloaded system.

We can't use such approach for our regular unit and integration tests because they are run on random machines including slow travis worker, under valgrind on with sanitizers that slow down execution, etc. These tests should be 100% reliable and thus their requirements should be less strict.

Asalle: > Please take a look at the related issue and discussions in the PR, in particular this one: roc-project/roc#283 (comment)

alright, i will take a look at more relaxed tests, thanks for the elaborated hints!

Asalle: I think something went wrong with github pages after rename, at list I'm not able to go to this link https://roc-streaming.org/roc/docs/internals/data_flow.html

gavv: The new URL is https://roc-streaming.org/toolkit/docs/internals/data_flow.html

(/roc/ was changed to /toolkit/)

Current website still have some broken URLs. I'll fix them during next few days. If you'll find broken URLs later than in a few day, please report them to me :)

roc/receiver.go
```diff
@@ -34,6 +36,44 @@ func OpenReceiver(rocContext *Context, receiverConfig *ReceiverConfig) (*Receive
...
+		return ErrInvalidArgs // roc_receiver_bind returns negative in 3 cases: invalid args, cannot bind address and not enough resources, maybe we should consider renaming this error to a more generic name?
+	}
+	panic(fmt.Sprintf(
+		"unexpected return code %d from roc_receiver_bind()", errCode))
+}
+
+func (r *Receiver) ReadFloats(floats []float32) error {
+	c_roc_frame := C.roc_frame{
+		samples:      (unsafe.Pointer(&floats[0])),
+		samples_size: (C.ulong)(len(floats) * 4),
```
gavv, line 58: Let's replace `4` with `unsafe.Sizeof(floats[0])`

roc/receiver.go
```diff
@@ -34,6 +36,44 @@ func OpenReceiver(rocContext *Context, receiverConfig *ReceiverConfig) (*Receive
...
+		return nil
+	}
+	if errCode < 0 {
+		return ErrInvalidArgs // roc_receiver_bind returns negative in 3 cases: invalid args, cannot bind address and not enough resources, maybe we should consider renaming this error to a more generic name?
+	}
+	panic(fmt.Sprintf(
+		"unexpected return code %d from roc_receiver_bind()", errCode))
+}
+
+func (r *Receiver) ReadFloats(floats []float32) error {
```
gavv, line 55: We should check if `floats` is an empty or nil slice. We can panic or silently return in this case. I think we should mimic io.Reader here, i.e. some of its common implementations.

gavv:
@Asalle what help do you need? :)

Asalle:
i patched up the test and semantically it should work fine, however it fails with a runtime error 
```
panic: runtime error: cgo argument has Go pointer to Go pointer

goroutine 20 [running]:
_/mnt/me/foss/roc-foss/roc-go/roc.(*Sender).WriteFloats.func1(0xbbc310, 0xc000098510, 0x1c66f4f32660)
	/mnt/me/foss/roc-foss/roc-go/roc/sender.go:74 +0x7f
_/mnt/me/foss/roc-foss/roc-go/roc.(*Sender).WriteFloats(0xbbc310, 0xc0000ae110, 0x2, 0x2, 0x6541c0, 0x300000000000000)
	/mnt/me/foss/roc-foss/roc-go/roc/sender.go:74 +0x70
_/mnt/me/foss/roc-foss/roc-go/roc.Test_roc_sender_write_receiver_read.func1(0xbbc310, 0xc0000ae110, 0x2, 0x2, 0xc0000d6120, 0xc00008e2a0)
	/mnt/me/foss/roc-foss/roc-go/roc/sender-receiver_test.go:102 +0x62
created by _/mnt/me/foss/roc-foss/roc-go/roc.Test_roc_sender_write_receiver_read
	/mnt/me/foss/roc-foss/roc-go/roc/sender-receiver_test.go:100 +0x10d
FAIL	_/mnt/me/foss/roc-foss/roc-go/roc	0.005s
FAIL
```
The deal is that we cannot pass a golang pointer to to c code, this is the place where it happens:
```
func (s *Sender) WriteFloats(frame []float32) error {
	c_roc_frame := C.roc_frame{
		samples:      unsafe.Pointer(&frame[0]),
		samples_size: (C.ulong)(len(frame) * 4),
	}

	errCode := C.roc_sender_write((*C.roc_sender)(s), &c_roc_frame)
       // ....
```
aka c_roc_frame contains a go pointer to c code, but when I try to cast it to `*C.void` it says that `samples` is supposed not to be cast. I'm currently investigating this, but if you can give me a hint, I would appreciate that @gavv

gavv:
Hm, after quick googling I've found this thread, did you see it? https://github.com/golang/go/issues/14210

Probably it's a cgo limitation.

We can try to create a jumper function in C (remember my logger PR?) that will accept roc_sender*, float*, and size_t. Internally it will construct roc_frame and call roc_sender_write. In other words, we'll move roc_frame construction from Go to C. This way we will pass float* directly to a С function (without wrapping it into C.roc_frame), probably this'll make cgo happy.

Asalle:
Yup, saw that briefly, but didn't get to the bottom of it.
Jumper function sounds fine.

gavv:
BTW it may make sense to start with only simple smoke tests in this PR (just call write and read and don't check anything, just ensure that it doesn't crash or hang), and merge it. More advanced tests may be implemented in further PRs.

Asalle:
I added a jumper funtion like so: 
```
/*
 #include <roc/sender.h>
int roc_sender_write_frame(roc_sender* sender, float* samples, unsigned long samples_size) {
    roc_frame frame = {(void*)samples, samples_size};
    return roc_sender_write(sender, &frame);
 }
*/
```
and I simplified a test for now to just write something, but it seems that somewhere on the way the memmory is corrupted, this is what I get when running the test
```
SIGABRT: abort
PC=0x7fd9e42b9a25 m=0 sigcode=18446744073709551610

goroutine 0 [idle]:
runtime: unknown pc 0x7fd9e42b9a25
stack: frame={sp:0x7ffcd8e29670, fp:0x0} stack=[0x7ffcd862a978,0x7ffcd8e299b0)
00007ffcd8e29570:  0000000000000000  00007fd9e4490d2b 
00007ffcd8e29580:  00007ffcd8e297c8  00007ffcd8e297e8 
00007ffcd8e29590:  00007fd9e44c96b0  0000ffff00001fa0 
00007ffcd8e295a0:  00007fd9e44c81c0  0000000000000000 
00007ffcd8e295b0:  00007fd9e44c8170  00007fd9e44c81c0 
00007ffcd8e295c0:  75662e646165725f  697013020031636e 
00007ffcd8e295d0:  0000000000000000  00000000ff000000 
00007ffcd8e295e0:  000000000000ac44  00007fd900000003 
00007ffcd8e295f0:  0000000011e1a300  0000000000000014 
00007ffcd8e29600:  0000000017d78400  000000003ba3d70a 
00007ffcd8e29610:  0000000000000064  000000003b9aca00 
00007ffcd8e29620:  0000000000000000  0000000000000000 
00007ffcd8e29630:  00007fd9e44c8170  00007fd9e44c81c0 
00007ffcd8e29640:  75662e646165725f  697013020031636e 
00007ffcd8e29650:  0000000000000000  00007fd9e4456d83 
00007ffcd8e29660:  0000000000000005  0000000000000001 
00007ffcd8e29670: <0000000000000000  00007ffcd8e296b0 
00007ffcd8e29680:  0000000117d78400  000000003ba3d70a 
00007ffcd8e29690:  00007fd9e4456c20  00007ffcd8e296b0 
00007ffcd8e296a0:  6f742074726f7000  0000000000000000 
00007ffcd8e296b0:  000000000000000b  000000000257cc10 
00007ffcd8e296c0:  000000000257cbe8  00007fd900000000 
00007ffcd8e296d0:  0000000000000000  e9eacc04fd29ba00 
00007ffcd8e296e0:  00007fd9e4271950  000000000257c6b0 
00007ffcd8e296f0:  fffffffe7fffffff  ffffffffffffffff 
00007ffcd8e29700:  ffffffffffffffff  ffffffffffffffff 
00007ffcd8e29710:  ffffffffffffffff  ffffffffffffffff 
00007ffcd8e29720:  ffffffffffffffff  ffffffffffffffff 
00007ffcd8e29730:  ffffffffffffffff  ffffffffffffffff 
00007ffcd8e29740:  ffffffffffffffff  ffffffffffffffff 
00007ffcd8e29750:  ffffffffffffffff  ffffffffffffffff 
00007ffcd8e29760:  ffffffffffffffff  ffffffffffffffff 
runtime: unknown pc 0x7fd9e42b9a25
stack: frame={sp:0x7ffcd8e29670, fp:0x0} stack=[0x7ffcd862a978,0x7ffcd8e299b0)
00007ffcd8e29570:  0000000000000000  00007fd9e4490d2b 
00007ffcd8e29580:  00007ffcd8e297c8  00007ffcd8e297e8 
00007ffcd8e29590:  00007fd9e44c96b0  0000ffff00001fa0 
00007ffcd8e295a0:  00007fd9e44c81c0  0000000000000000 
00007ffcd8e295b0:  00007fd9e44c8170  00007fd9e44c81c0 
00007ffcd8e295c0:  75662e646165725f  697013020031636e 
00007ffcd8e295d0:  0000000000000000  00000000ff000000 
00007ffcd8e295e0:  000000000000ac44  00007fd900000003 
00007ffcd8e295f0:  0000000011e1a300  0000000000000014 
00007ffcd8e29600:  0000000017d78400  000000003ba3d70a 
00007ffcd8e29610:  0000000000000064  000000003b9aca00 
00007ffcd8e29620:  0000000000000000  0000000000000000 
00007ffcd8e29630:  00007fd9e44c8170  00007fd9e44c81c0 
00007ffcd8e29640:  75662e646165725f  697013020031636e 
00007ffcd8e29650:  0000000000000000  00007fd9e4456d83 
00007ffcd8e29660:  0000000000000005  0000000000000001 
00007ffcd8e29670: <0000000000000000  00007ffcd8e296b0 
00007ffcd8e29680:  0000000117d78400  000000003ba3d70a 
00007ffcd8e29690:  00007fd9e4456c20  00007ffcd8e296b0 
00007ffcd8e296a0:  6f742074726f7000  0000000000000000 
00007ffcd8e296b0:  000000000000000b  000000000257cc10 
00007ffcd8e296c0:  000000000257cbe8  00007fd900000000 
00007ffcd8e296d0:  0000000000000000  e9eacc04fd29ba00 
00007ffcd8e296e0:  00007fd9e4271950  000000000257c6b0 
00007ffcd8e296f0:  fffffffe7fffffff  ffffffffffffffff 
00007ffcd8e29700:  ffffffffffffffff  ffffffffffffffff 
00007ffcd8e29710:  ffffffffffffffff  ffffffffffffffff 
00007ffcd8e29720:  ffffffffffffffff  ffffffffffffffff 
00007ffcd8e29730:  ffffffffffffffff  ffffffffffffffff 
00007ffcd8e29740:  ffffffffffffffff  ffffffffffffffff 
00007ffcd8e29750:  ffffffffffffffff  ffffffffffffffff 
00007ffcd8e29760:  ffffffffffffffff  ffffffffffffffff 

goroutine 6 [syscall]:
runtime.cgocall(0x505960, 0xc000044688, 0xc000000000)
	/usr/lib/golang/src/runtime/cgocall.go:133 +0x5b fp=0xc000044658 sp=0xc000044620 pc=0x4055db
_/mnt/me/foss/roc-foss/roc-go/roc._Cfunc_roc_sender_close(0x257e310, 0x0)
	_cgo_gotypes.go:348 +0x49 fp=0xc000044688 sp=0xc000044658 pc=0x502d69
_/mnt/me/foss/roc-foss/roc-go/roc.(*Sender).Close.func1(0x257e310, 0x0)
	/mnt/me/foss/roc-foss/roc-go/roc/sender.go:86 +0x4d fp=0xc0000446b8 sp=0xc000044688 pc=0x504ced
_/mnt/me/foss/roc-foss/roc-go/roc.(*Sender).Close(0x257e310, 0x0, 0x0)
	/mnt/me/foss/roc-foss/roc-go/roc/sender.go:86 +0x2f fp=0xc000044710 sp=0xc0000446b8 pc=0x50421f
_/mnt/me/foss/roc-foss/roc-go/roc.Test_roc_sender_write_receiver_read(0xc000128120)
	/mnt/me/foss/roc-foss/roc-go/roc/sender-receiver_test.go:143 +0x126 fp=0xc000044780 sp=0xc000044710 pc=0x502436
testing.tRunner(0xc000128120, 0x558bf0)
	/usr/lib/golang/src/testing/testing.go:991 +0xdc fp=0xc0000447d0 sp=0xc000044780 pc=0x4c453c
runtime.goexit()
	/usr/lib/golang/src/runtime/asm_amd64.s:1373 +0x1 fp=0xc0000447d8 sp=0xc0000447d0 pc=0x466001
created by testing.(*T).Run
	/usr/lib/golang/src/testing/testing.go:1042 +0x357

goroutine 1 [chan receive]:
testing.(*T).Run(0xc000128120, 0x555827, 0x23, 0x558bf0, 0x49fd01)
	/usr/lib/golang/src/testing/testing.go:1043 +0x37e
testing.runTests.func1(0xc000128000)
	/usr/lib/golang/src/testing/testing.go:1284 +0x78
testing.tRunner(0xc000128000, 0xc000055df0)
	/usr/lib/golang/src/testing/testing.go:991 +0xdc
testing.runTests(0xc00000e080, 0x63f340, 0x7, 0x7, 0x0)
	/usr/lib/golang/src/testing/testing.go:1282 +0x2a7
testing.(*M).Run(0xc00011e000, 0x0)
	/usr/lib/golang/src/testing/testing.go:1199 +0x15f
_/mnt/me/foss/roc-foss/roc-go/roc.TestMain(0xc00011e000)
	/mnt/me/foss/roc-foss/roc-go/roc/utils_test.go:18 +0x47
main.main()
	_testmain.go:54 +0x135

goroutine 7 [runnable]:
_/mnt/me/foss/roc-foss/roc-go/roc._Cfunc_roc_sender_write_frame(0x257e310, 0xc00001e1f8, 0x8, 0x0)
	_cgo_gotypes.go:391 +0x4d
_/mnt/me/foss/roc-foss/roc-go/roc.(*Sender).WriteFloats.func1(0x257e310, 0xc00001e1f8, 0x2, 0x2, 0x4, 0xc00001c360)
	/mnt/me/foss/roc-foss/roc-go/roc/sender.go:74 +0x74
_/mnt/me/foss/roc-foss/roc-go/roc.(*Sender).WriteFloats(0x257e310, 0xc00001e1f8, 0x2, 0x2, 0x0, 0x0)
	/mnt/me/foss/roc-foss/roc-go/roc/sender.go:74 +0x56
_/mnt/me/foss/roc-foss/roc-go/roc.Test_roc_sender_write_receiver_read.func1(0x257e310, 0xc00001e1f8, 0x2, 0x2, 0xc000128120, 0xc00001c360)
	/mnt/me/foss/roc-foss/roc-go/roc/sender-receiver_test.go:102 +0x62
created by _/mnt/me/foss/roc-foss/roc-go/roc.Test_roc_sender_write_receiver_read
	/mnt/me/foss/roc-foss/roc-go/roc/sender-receiver_test.go:100 +0x102

rax    0x0
rbx    0x7fd9e3ee3380
rcx    0x7fd9e42b9a25
rdx    0x0
rdi    0x2
rsi    0x7ffcd8e29670
rbp    0xc000044648
rsp    0x7ffcd8e29670
r8     0x0
r9     0x7ffcd8e29670
r10    0x8
r11    0x246
r12    0xc000044800
r13    0x0
r14    0x257c4b0
r15    0x0
rip    0x7fd9e42b9a25
rflags 0x246
cs     0x33
fs     0x0
gs     0x0
FAIL	_/mnt/me/foss/roc-foss/roc-go/roc	0.007s
FAIL
```
I'm currently investigating this, any help appreciated.

gavv:
Here is how to debug such things:

```
$ go test -c
$ ls -l roc.test
-rwxr-xr-x 1 victor victor 3408224 Aug  3 13:40 roc.test
$ gdb ./roc.test
....
Thread 1 "roc.test" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7d0b55b in __GI_abort () at abort.c:79
#2  0x00007ffff7cc382e in ?? () from /usr/lib64/libuv.so.1
#3  0x00007ffff7ef74b5 in roc::core::Mutex::~Mutex (this=0x696ea8) at src/modules/roc_core/target_libuv/roc_core/mutex.h:40
#4  0x00007ffff7efb3b6 in roc_sender::~roc_sender (this=0x696d30) at src/lib/src/private.h:59
#5  0x00007ffff7efb201 in roc::core::IAllocator::destroy<roc_sender> (this=0x695190, object=...) at src/modules/roc_core/iallocator.h:38
#6  0x00007ffff7efb01c in roc_sender_close (sender=0x696d30) at src/lib/src/sender.cpp:312
#7  0x0000000000505cc7 in _cgo_5961a802b2e7_Cfunc_roc_sender_close (v=0xc00003c688) at cgo-gcc-prolog:73
#8  0x0000000000465950 in runtime.asmcgocall () at /usr/lib/go/src/runtime/asm_amd64.s:655
#9  0x000000000046356f in runtime.newproc.func1 () at /usr/lib/go/src/runtime/proc.go:3387
#10 0x0000000000464176 in runtime.systemstack () at /usr/lib/go/src/runtime/asm_amd64.s:370
#11 0x00000000004397a0 in ?? () at <autogenerated>:1
#12 0x0000000000464004 in runtime.rt0_go () at /usr/lib/go/src/runtime/asm_amd64.s:220
#13 0x0000000000506620 in crosscall_amd64 ()
#14 0x000000000046400b in runtime.rt0_go () at /usr/lib/go/src/runtime/asm_amd64.s:225
#15 0x0000000000000001 in ?? ()
#16 0x00007fffffffdd38 in ?? ()
#17 0x0000000000000001 in ?? ()
#18 0x00007fffffffdd38 in ?? ()
#19 0x0000000000000000 in ?? ()
```

So abort() is called in Mutex::~Mutex(). Go to its source code: https://github.com/roc-streaming/roc-toolkit/blob/master/src/modules/roc_core/target_libuv/roc_core/mutex.h#L40

It just calls uv_mutex_destroy(). Now go to libuv sources: https://github.com/libuv/libuv/blob/v1.x/src/unix/thread.c#L326

It aborts if pthread_mutex_destroy() returns error.

Now go to the man page: https://linux.die.net/man/3/pthread_mutex_destroy

```
Errors

The pthread_mutex_destroy() function may fail if:

EBUSY
    The implementation has detected an attempt to destroy the object referenced by mutex while it is locked or referenced (for example, while being used in a pthread_cond_timedwait() or pthread_cond_wait()) by another thread. 
EINVAL
    The value specified by mutex is invalid.
```

So we're either passing an invalid address instead of mutex, or trying to destroy a mutex still being locked.

The second can happen if you call sender's close() and write() concurrently, for example. Now we go to your test, and see that this is exactly what happening. We're starting goroutine that calls write() in a loop and then call close() without waiting until the goroutine finishes.

So you just need to fix the test.

I think it would be helpful to add simple "race detection" for such cases, and report a meaningful error when trying to close sender/receiver being used. Added to my TODO.

roc/sender.go
```diff
@@ -1,9 +1,95 @@
 package roc
 
 /*
-#include <roc/sender.h>
+ #include <roc/sender.h>
+int roc_sender_write_frame(roc_sender* sender, float* samples, unsigned long samples_size) {
```
gavv, line 5: Let's rename it to something like `rocGoSenderWriteFrame`, similar to https://github.com/roc-streaming/roc-go/blob/master/roc/log.c#L4

Or, better, `rocGoSenderWriteFloats`.

roc/sender.go
```diff
@@ -1,9 +1,95 @@
...
+	if errCode < 0 {
+		return ErrInvalidArgs // roc_receiver_bind returns negative in 3 cases: invalid args, cannot bind address and not enough resources, maybe we should consider renaming this error to a more generic name?
+	}
+	panic(fmt.Sprintf(
+		"unexpected return code %d from roc_address_init()", errCode))
+}
+
+func (s *Sender) WriteFloats(frame []float32) error {
+	sizeOfFloat := (int)(unsafe.Sizeof(frame[0]))
+	errCode := C.roc_sender_write_frame((*C.roc_sender)(s), (*C.float)(unsafe.Pointer(&frame[0])), (C.ulong)(len(frame)*sizeOfFloat))
```
gavv, line 74: Consider moving sizeof into the jumper function (since it's already knows the sample type (float)).

Also, the unsafe.Pointer here is unnecessary.

roc/sender-receiver_test.go
```diff
@@ -0,0 +1,151 @@
...
+		t.Fail()
+	}
+	err = sender.Close()
+	if err != nil {
+		t.Fail()
+	}
+	err = ctx.Close() // remove after finalizers are done
+	if err != nil {
+		t.Fail()
+	}
```
gavv, line 150: Maybe extract this into function too, similar to prepare()?

Also, maybe it makes sense to create a struct ("test fixture") with these triplet (sender, receiver, context).

gavv:
Another way to debug your problem would be bisecting via disabling or commenting out parts of code. In this case, you can disable first write, all writes, and calls to close, and see if the test is still failing. This way you can understand that close() is the reason of the failure.

Also, just enabling debug or trace logs in libroc may help:

```
=== RUN   Test_roc_sender_write_receiver_read
transceiver: removing port 127.0.0.1:45677
sender: using remote source port rtp+rs8m:127.0.0.1:45677
udp receiver: closing port 127.0.0.1:45677
sender: using remote repair port rs8m:127.0.0.1:38865
udp receiver: closed port 127.0.0.1:45677
transceiver: asynchronous close finished: port 127.0.0.1:45677
of encoder: initializing: codec=rs m=8
fec writer: update block size: cur_sbl=0 cur_rbl=0 new_sbl=20 new_rbl=10
packetizer: initializing: n_channels=2 samples_per_packet=309
transceiver: removing port 127.0.0.1:38865
udp receiver: closing port 127.0.0.1:38865
udp receiver: closed port 127.0.0.1:38865
transceiver: asynchronous close finished: port 127.0.0.1:38865
roc_receiver: closed receiver
transceiver: removing port 127.0.0.1:50067
udp sender: closing port 127.0.0.1:50067
udp receiver: closed port 127.0.0.1:50067
transceiver: asynchronous close finished: port 127.0.0.1:50067
SIGABRT: abort
PC=0x7f572a5e6861 m=0 sigcode=18446744073709551610

goroutine 0 [idle]:
runtime: unknown pc 0x7f572a5e6861
```

This output also can give you a hint that the crash happens after calling close.

roc/receiver.go
Asalle: not sure which behaviour io.Reader has, but I return ErrInvalidArgs if floats is nil

Asalle:
@gavv I'm not sure how code coverage could decrease if I added a big test with this pr :thinking: 
but I'll try to add one more unit test somewhere

Asalle:
I don't think I can add any more unit tests

gavv:
@Asalle Hi, sorry for delay, I'm back. Will review the changes now.

> btw, can we somehow add travis check per pr? I don't think I have the permissions to add that

They're enabled, but travis sometimes fails to link them to PR. I think it's a bug in travis, I see it many times. In this case you can find them manually here: https://travis-ci.org/github/roc-streaming/roc-go/pull_requests

> I'm not sure how code coverage could decrease if I added a big test with this pr

I think it decreased because we also added new lines of the source code which are not covered. You can find uncovered lines by clicking file name: https://coveralls.io/builds/33263891

They're mostly various error paths. I think it's OK to keep them uncovered, at least for now.

gavv:
BTW please don't resolve discussions by yourself, keep them for reviewer. Otherwise it's hard to tell what's need to be re-reviewed since last review. To mark them as done, you can leave a comment or thumbs up.

roc/receiver.go
```diff
@@ -34,6 +39,38 @@ func OpenReceiver(rocContext *Context, receiverConfig *ReceiverConfig) (*Receive
...
+	}
+	panic(fmt.Sprintf(
+		"unexpected return code %d from roc_receiver_bind()", errCode))
+}
+
+func (r *Receiver) ReadFloats(frame []float32) error {
+	if frame == nil {
+		return ErrInvalidArgs
+	}
+	errCode := C.rocGoReceiverReadFloats((*C.roc_receiver)(r), (*C.float)(&frame[0]), (C.ulong)(len(frame)))
```
gavv, line 62: `frame[0]` will panic if frame is not nil, but its len is zero. So we should also check `if len(frame) == 0`. I think in this case we can just do nothing return nil (no error). Same for WriteFloats.

gavv (commented):
The changes look great! Only two minor issues.

gavv:
Merged & fixed: https://github.com/roc-streaming/roc-go/commit/eb86104863e1f6566c9db88a31e2e65d19fcb72b

gavv:
@Asalle Thanks!!

Issue #11, opened 2019-11-13
Add API documentation


gavv:
We should add documentation for the package and for all public functions. Most docs can be copied from libroc headers.

- [x] Endpoint
- [x] Config
- [x] Config
- [x] Context
- [x] Sender
- [x] Receiver
- [x] Log

gavv:
I'll finish this after merging #24.

gavv:
Unassigning this from myself for now. I'll pick it up again after a few weeks if it will still be open.

gavv:
It makes sense to return to it after #30, because docs in 0.2 are changed too.

PR #25, opened 2020-06-26
Address documentation and errors cleanup

gavv:
See commit messages for details.

I recommend to review each commit individually.

#11

### DIFF
--- a/roc/address.go
+++ b/roc/address.go
@@ -10,13 +10,28 @@ import (
 	"unsafe"
 )
 
-// Address as declared in roc/address.h:59
+// Address represents network endpoint address.
+//
+// Consists of IP address plus UDP or TCP port number.
+// Similar to net.Addr in Go and struct sockaddr in C.
+//
+// Should not be used concurrently
 type Address struct {
 	raw *C.roc_address
 	mem []byte
 }
 
-// NewAddress parses the `ip`, `port` and `family` and initializes the Address object
+// NewAddress creates and initializes a new Address.
+//
+// The IP address is parsed from a string representation. If family is AfAuto, the
+// address family is auto-detected from string format. Otherwise, the string format
+// should correspond to the family specified.
+//
+// The port number should be in range [0; 65536).
+//
+// When Address is used to bind a sender or receiver port, the "0.0.0.0" (for IPv4)
+// or "::" (for IPv6) may be used to bind the port to all network interfaces, and
+// zero port number may be used to bind the port to a randomly chosen ephemeral port.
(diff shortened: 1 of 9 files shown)


gavv:
https://travis-ci.org/github/roc-project/roc-go/builds/702301867

gavv:
@Asalle what do you think about 3469532ea454656db27854677a5af1f9cd95a598?

I'm not 100% sure, but I feel that this approach (panics instead of defensive checks for everything) is more familiar in Go world. One though supporting it is that panics in Go are less "fatal" than in C (they can be worked around using recover).

roc/address.go
```diff
@@ -48,44 +49,46 @@ func NewAddress(family Family, ip string, port int) (*Address, error) {
...
 //
 // If AfAuto was used to construct address, the actually selected family, i.e.
 // either AfIPv4 or AfIPv6, is reported.
-func (a *Address) Family() (Family, error) {
+func (a *Address) Family() Family {
 	f := C.roc_address_family(a.raw)
 	family := (Family)(f)
 	if family == afInvalid {
-		return family, ErrInvalidArguments
+		panic("unexpected failure in roc_address_family()")
```
Asalle, line 65: I don't think it's a good idea to panic here, it  will stop the work of the whole app afaik
the right thing to do here as a library is to report and error and move on
I do agree that in line 53 we should panic, because this situation when API return an unknown code is not possible and signalises that we don't follow the contract

gavv: Well if we can panic on line 53, I think we can also panic here. roc_address_family() can fail (and return AF_INVALID) only if passed `address` is NULL or was not successfully initialized.

In the C API, the API itself allows the user to make such calls: the user can just pass NULL or uninitialized memory. As the API allows to do it, we handle it, and report error instead of blindly crashing.

However in the Go API, such case is not possible: NewAddress() will either fail or return non-null and fully initialized address object. So if the user has an object on which Family() can be called, Family() is guaranteed to succeed.

BTW I made `afInvalid` unexported for the same reason: it should never be visible by the user of Go API.

So if roc_address_family() fails here, it means that there is a bug either in C or Go part, just like on line 53.

So we should decide whether to panic on such "impossible" bugs detected in Go and C API. If you think the panic on line 53 is okay, then I guess this one is okay too.

Asalle: this doesn't feel right to me, but you have a point, address should be fully initialized
I think we can keep it like this and see what's going to come out of it

gavv: OK, so for now we'll use the following approach: if we detect a violation of the C API contract, or violation of the internal invariant in Go bindings, we panic.

PR #32, opened 2021-02-16
Switch to github actions

gavv:
This PR removes travis config and adds github actions config.

travis-ci.org is shutting down, users are suggested to migrate to travis-ci.com. I tried, and I even didn't finish migration of all repos yet and already exhausted monthly limits :) So it seems we should migrate all or some of our repos somewhere. Github actions are an obvious choice, since they're integrated into github and AFAIK have no limits for OSS projects so far.


gavv:
How it will look like: https://github.com/gavv/roc-go/actions/runs/571380831

gavv:
> I even didn't finish migration of all repos yet and already exhausted monthly limits :)

Update: it wasn't actually a monthly limit, it was an overall limit (not renewable). Renewable free minutes for OSS are only by request now. I requested some minutes for roc-java (which is harder to migrate from travis), no response for now. After merging this PR, all repos except roc-java will be migrated to GH actions.

gavv:
@Asalle ping! :)

gavv:
@Asalle I'll merge both PRs, but feel free to add review comments later if you'll have some.

PR #39, opened 2022-12-23
Add FUNDING.yml

gavv:
See https://github.com/roc-streaming/roc-toolkit/discussions/395#discussioncomment-4475754

@Asalle feel free to join liberapay and open-collective teams if you would like to receive donations!


Asalle:
@gavv thanks! I don't feel like I'm doing enough to get any money for it so far, but maybe in future! :)

Issue #33, opened 2021-03-03
Add quick start section to README


gavv:
Add quick start section with code snippets with sender and receiver.

Issue #4, opened 2019-11-09
Take headers and lib from pkg-config 


Asalle:
Currently we have hardcoded `/usr/lib/roc` path that is not always where you have your headers/lib. Let's take that info from pkg-config.

gavv:
I've updated https://github.com/roc-project/roc/issues/271

Asalle:
Can you add a label "blocked" or "camping"? Or can you give me the permissions to add tags? :3

gavv:
Hm, I'm don't see such a permission in settings.

Probably it's better to user projects for that? Like we do it in core: https://github.com/roc-project/roc/projects/2

I've enabled projects for the repo, you can try it (do you have permissions?). Or I can just create the label, I'm OK with that too.

Asalle:
Have no permissions to add a label, also I don't see the board. But no
worries, I don't need it too much!

gavv:
> Have no permissions to add a label, also I don't see the board.

Not good. I've selected a different permission level (triage -> write) for the go team, could you check now?

Asalle:
works! thanks

gavv:
.pc generation & installation landed in develop branch. `scons install` will now install `roc.pc`.

cc @fusuiyi123

Asalle:
Yay! We can start using it in go bindings!

Asalle:
now that I've looked into it, is this feature relevant? The only place where we use the C includes are comments like 
```
/*
#include <roc/address.h>
*/
import "C"
```
which work regardless if one uses pkg-config or just installed them with their distribution's package manager.

gavv:
If the user installed roc into a prefix which is not in default include or library search path list, compilation will fail. pkg-config will solve this issue.

A real-world examples. When you're using SoX library you should write:

```
#include <sox.h>
```

On most distros, sox.h is located at /usr/include/sox.h and include works without any pkg-config.

However, on some CentOS version, for some reason maintainers decided to install that file at /usr/include/sox/sox.h. On those CentOS systems pkg-config for SoX returns `-I/usr/include/sox`. If you're using pkg-config, everything works, but without it your include fails.

So yes, we still need pkg-config, though it's not critical.

Note that pkg-config generation is currently available only in develop, while bindings are using master.

gavv:
Example of using pkg-config with cgo: https://github.com/aam335/speexdsp/blob/master/speexdsp.go#L4

Asalle:
[...]

PR #40, opened 2022-12-23
Update to 0.2

### LINKED ISSUE #30: Mirgrate the bindings to roc-toolkit 0.2

gavv:
Closes #30, #11, #4, #33.

Changes:

* update bindings to new API from 0.2; this includes replacement of roc_address with roc_endpoint; Endpoint is a regular Go struct without native part; native roc_endpoint is created only temporary when it's needed; this way user should not bother to manually deinitialize endpoint, which would be inconvenient

* port comments from C lib to go; now go bindings are completely documented

* take cflags and ldflags from pkg-config

* remove ErrInvalidArgs; instead, add nativeErr, which holds error code from C lib; in 0.2 all libroc functions return error codes (but most time just -1); accordingly, in go we now always return nativeError with that code; since libroc does not publish error codes yet, nativeErr is not exported for now

* ensure that subsequent Close call is no-op (otherwise it would cause use-after-free in native code); to achieve this and keep code thread-safe, I had to add rwlocks; all operations are guarded with readlock to allow parallelism, and close operation is guarded with write lock

* update tests; I employed testify in some of them, hope you don't mind

* add exhaustruct linter and enable it for `roc_` structs; it checks that every field is initialized; it allows us to be sure that we don't forget fields in structs like `C.struct_roc_sender_config`

* run tests with cgocheck=2 (it checks how memory is passed from go to C)

* update CI to build with 0.2

* cleanup Makefile and README


gavv:
Update: added quick start to README https://github.com/roc-streaming/roc-go/blob/4ac7c8d66511eaba82f333e0ed1946d4bc6d8d37/README.md#quick-start

gavv:
@Asalle hi, I think of posting an announcement and tutorial for 0.2 in upcoming week or two, do you plan to look at it relatively soon? In this case I'll wait to include go bindings in too.

gavv:
No worries if you don't have time right now, we can release bindings later. I'm just sorting out my todo.

Makefile
```diff
@@ -1,22 +1,22 @@
 GO111MODULE := on
 export GO111MODULE
 
-all: check test race
+all: build lint test
```
Asalle, line 4: nice that now `test` runs `race` by default! Would be nice to add a `clean` option, which is trivial but idiomatically aligns with the rest of the targets

roc/config.go
```diff
@@ -5,68 +5,238 @@ package roc
...
+
+	// RTCP over UDP (RFC 3550).
+	//
+	// Interfaces:
+	//  - InterfaceAudioControl
+	//
+	// Transports:
+	//  - UDP
+	///
+	ProtoRrcp Protocol = 70
```
Asalle, line 211: just a question - are these consts taken from the c code? I'm wondering what's the logic behind the scattered values

roc/context_test.go
```diff
@@ -1,36 +1,38 @@
...
 
-		err = ctx.Close()
-		if err != nil {
-			fail(nil, err, t)
-		}
+				err = ctx.Close()
+				require.NoError(t, err)
+			} else {
+				require.Error(t, err)
+				require.Nil(t, ctx)
```
Asalle, line 34: I think checking for a specific error provides for more clarity. Resorting to general errors with a string inside should be the case only if the error is unexpected. This way we can handle different types of errors and completely panic only in this case of unexpected unidentified error.

Asalle (commented):
Thanks for this huge PR! Amazing work and sorry for this late review. I'm trying to implement a new schedule for open source, hopefully it will allow me to reply with max 1 week of delay.

Makefile
gavv: Done

roc/context_test.go
gavv: Good point! Reworked all tests to check for specific errors.

roc/log.go
```diff
@@ -16,32 +17,57 @@ import (
...
+// LogMessage defines message written to log.
+type LogMessage struct {
+	// Message log level.
+	Level LogLevel
+
+	// Name of the module that originated the message.
+	Module string
+
+	// Name of the source code file.
+	// May be emoty.
```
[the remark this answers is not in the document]
gavv: Fixed

roc/config.go
gavv: 1. Yep, numbers come from C API.

2. They are scattered because I think new values will be added in between, and it would be nice to keep their numeric values grouped. For example, aggregate protocols will be 10-20, transport protocols will be 20-30, FEC source and repair protocols start from 30, etc.

3. In future it would be nice to create a script that either automatically updates config.go file from C header or maybe just verifies that they're in-sync and reports what is missing / different. I don't mean a generic binding generator, just some simple solution specific to this given file. Let me know if you have ideas how we could do it...

gavv: Another thing that bothers me is that all these comments can quickly become outdated compared to the C API. Not sure what we could do here.

gavv:
PR is ready for review again.

gavv:
Gentle ping :)

Asalle:
make test passes locally with roc-toolkit 0.2.0

Issue #45, opened 2022-12-23
Add unit test for convert functions


gavv:
Cover functions from convert.go with unit tests.

gavv:
Landed

PR #49, opened 2023-02-27
add tests convert functions

ortex:
#45

### DIFF
--- a/roc/convert.go
+++ b/roc/convert.go
@@ -4,6 +4,8 @@ import (
 	"C"
 )
 
+type char = C.char
+
 func go2cStr(str string) []C.char {
 	charArray := make([]C.char, len(str)+1)
 	for ind, r := range str {
--- a/roc/convert.go
+++ b/roc/convert.go
@@ -6,16 +6,16 @@ import (
 
 type char = C.char
 
-func go2cStr(str string) []C.char {
-	charArray := make([]C.char, len(str)+1)
+func go2cStr(str string) []char {
+	charArray := make([]char, len(str)+1)
 	for ind, r := range str {
-		charArray[ind] = (C.char)(r)
+		charArray[ind] = (char)(r)
 	}
 	charArray[len(str)] = '\x00'
 	return charArray
 }
 
-func c2goStr(charArray []C.char) string {
+func c2goStr(charArray []char) string {
 	byteArray := make([]byte, 0, len(charArray)-1)
 	for _, c := range charArray {
 		if c == '\x00' {
--- a/roc/convert_test.go
+++ b/roc/convert_test.go
@@ -22,20 +22,35 @@ func Test_go2cBool(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


roc/convert_test.go
```diff
@@ -0,0 +1,41 @@
...
+		{false, 0},
+	}
+	for _, tt := range tests {
+		t.Run(strconv.FormatBool(tt.arg), func(t *testing.T) {
+			assert.Equal(t, tt.result, uint(go2cBool(tt.arg)))
+		})
+	}
+}
+
+func Test_go2cStr_c2goStr(t *testing.T) {
```
gavv, line 25: Could we also add tests for just c2goStr and just go2cStr?

The current tests don't actually check zero-byte handling, it would be nice to cover that part more precisely.

gavv (commented):
Thanks for PR!

roc/convert_test.go
ortex: added

roc/convert.go
```diff
@@ -4,6 +4,8 @@ import (
 	"C"
 )
 
+type char = C.char
```
gavv, line 7: nit: let's use `char` instead of `C.char` below

roc/convert_test.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func Test_c2goStr(t *testing.T) {
+	tests := []struct {
+		name   string
+		arg    []char
+		result string
+	}{
+		{name: "str", arg: []char{'s', 't', 'r', '\x00'}, result: "str"},
+		{name: "empty", arg: []char{'\x00'}, result: ""},
```
gavv, line 49: Let's add two more tests:

* two zero bytes at the end instead of one
* unexpected zero byte in the middle of the string

roc/convert_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+func Test_go2cStr(t *testing.T) {
+	tests := []struct {
+		name   string
+		arg    string
+		result []char
+	}{
+		{name: "str", arg: "str", result: []char{'s', 't', 'r', '\x00'}},
+		{name: "str00str", arg: "str\x00", result: []char{'s', 't', 'r', '\x00', '\x00'}},
+		{name: "empty", arg: "", result: []char{'\x00'}},
```
gavv, line 33: * let's rename `name: "str00str"` to `name: "str0str"` because there is only one zero byte
* let's add a test when zero byte is in the end

gavv:
https://github.com/roc-streaming/roc-go/issues/50

roc/convert.go
ortex: done

roc/convert_test.go
ortex: done

roc/convert_test.go
ortex: done

gavv:
Thanks!

PR #55, opened 2023-03-06
add error handling in go2cStr

### LINKED ISSUE #50: Add error handling to go2cStr
Currently, if go2cStr meets zero byte in the middle or in the end of the string, it stops early (and also oddly produces two zero bytes instead of one).

Instead, it should report error that the input string is incorrect. Everywhere where we call go2cStr, we should check error and report it to user.

PiotrLewandowski323:
fixes #50

### DIFF
--- a/roc/convert.go
+++ b/roc/convert.go
@@ -2,17 +2,22 @@ package roc
 
 import (
 	"C"
+	"fmt"
 )
 
 type char = C.char
 
-func go2cStr(str string) []char {
+func go2cStr(str string) ([]char, error) {
 	charArray := make([]char, len(str)+1)
 	for ind, r := range str {
-		charArray[ind] = (char)(r)
+		c := (char)(r)
+		if c == '\x00' {
+			return nil, fmt.Errorf("unexpected zero byte in the string: %q", str)
+		}
+		charArray[ind] = c
 	}
 	charArray[len(str)] = '\x00'
-	return charArray
+	return charArray, nil
 }
 
 func c2goStr(charArray []char) string {
--- a/roc/endpoint.go
+++ b/roc/endpoint.go
@@ -92,7 +92,10 @@ func ParseEndpoint(uri string) (*Endpoint, error) {
 		}
 	}()
 
-	cURI := go2cStr(uri)
+	cURI, err := go2cStr(uri)
+	if err != nil {
+		return nil, fmt.Errorf("invalid uri: %w", err)
+	}
 	errCode = C.roc_endpoint_set_uri(cEndp, (*C.char)(&cURI[0]))
 	if errCode != 0 {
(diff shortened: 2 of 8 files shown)


ortex:
LGTM. But, it would be great to add tests for endpoint, to cover new branches

PiotrLewandowski323:
The linux build in CI uses go 1.12 which doesn't support "%w" to wrap errors, so I had to replace it with "%v".

gavv:
I think we can bump minimum go to 1.13 (in go mod and .github/workflows). The choice of 1.12 with rather arbitrary (it just was known to work on 1.12). On the other hand, I'd say `%w` is a very very useful feature for error handling.

roc/endpoint.go
```diff
@@ -92,7 +92,10 @@ func ParseEndpoint(uri string) (*Endpoint, error) {
 		}
 	}()
 
-	cURI := go2cStr(uri)
+	cURI, err := go2cStr(uri)
+	if err != nil {
+		return nil, fmt.Errorf("invalid URI: %w", err)
```
gavv, line 97: nit: URI => uri (because we use lowercase in all other errors)

roc/convert.go
```diff
@@ -2,17 +2,22 @@ package roc
...
 type char = C.char
 
-func go2cStr(str string) []char {
+func go2cStr(str string) ([]char, error) {
 	charArray := make([]char, len(str)+1)
 	for ind, r := range str {
-		charArray[ind] = (char)(r)
+		c := (char)(r)
+		if c == '\x00' {
+			return nil, fmt.Errorf("unexpected zero byte in the string: %s", str)
```
gavv, line 15: net: let's use %q instead of %s for more readable error message

roc/endpoint_test.go
```diff
@@ -102,3 +103,58 @@ func TestEndpoint(t *testing.T) {
...
+		uri := "rtsp://192.168.0.1:12345\x00"
+		parseErr := errors.New("invalid URI: ")
+		endp, err := ParseEndpoint(uri)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), parseErr.Error())
+		require.Nil(t, endp)
+	})
+}
+
+func Test_UriError(t *testing.T) {
```
gavv, line 118: I think instead of adding to standalone tests TestEndpointErrors and Test_UriError we can just add cases to the table in TestEndpoint.

I suppose we'll need to modify TestEndpoint to use require.Contains for error checking instead of require.Equal.

gavv (commented):
Thanks for update! A couple more (I think final) comments.

PiotrLewandowski323:
I rebased onto main because endpoint tests were refactored and updated the changes according to your comments.

ortex (approved):

PR #56, opened 2023-03-07
Minor refactoring in endpoint tests

gavv:
Reorder, add comments, and commented test (see comments).


gavv:
https://github.com/roc-streaming/roc-toolkit/issues/519 is fixed, after releasing 0.2.3 we can uncomment the test

ortex (approved):

Issue #54, opened 2023-03-06
Add test for context closing


gavv:
Add test which checks that if there are receiver or sender attached to context, Context.Close returns error; and when there are no more receivers and senders, context can be successfully closed.

gavv:
@Darrellbor feel free to leave a comment so that I could assign you.

Darrellbor:
I would love to pick it up

gavv:
Landed!

PR #58, opened 2023-03-09
Tests for closing context

Darrellbor:
PR for #54 
implemented 
- over arching test for context closing 
- sub tests to handle cases like 
1. when there are no senders and no receivers
2. when there are receivers
3. when there are senders
4. when there are senders and receivers

### DIFF
--- a/roc/context_test.go
+++ b/roc/context_test.go
@@ -90,7 +90,7 @@ func TestContext_Close(t *testing.T) {
 					if err != nil {
 						t.Fail()
 					}
-					err = ctx.Close()
+					ctx.Close()
 				}()
 			}
 			if tt.hasSenders {
@@ -110,7 +110,7 @@ func TestContext_Close(t *testing.T) {
 					if err != nil {
 						t.Fail()
 					}
-					err = ctx.Close()
+					ctx.Close()
 				}()
 			}
 
--- a/roc/context_test.go
+++ b/roc/context_test.go
@@ -90,7 +90,10 @@ func TestContext_Close(t *testing.T) {
 					if err != nil {
 						t.Fail()
 					}
-					ctx.Close()
+
+					err = ctx.Close()
+					require.NoError(t, err)
+
 				}()
 			}
 			if tt.hasSenders {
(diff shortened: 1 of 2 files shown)


Darrellbor:
@gavv PR ready for review

ortex:
Thanks for PR!

Could you please also add a check that context closes successfully after closing the attached receiver/sender for these cases?

>when there are receivers
>when there are senders
>when there are senders and receivers

Darrellbor:
@ortex I've been looking into that, I just pushed the closing methods. they race condition test still fails locally

Darrellbor:
@ortex never mind, I figured out the issue that was causing the race condition.

ortex:
Now tests actually don't assert that context closing successfully after closing the attached receiver or sender.

And about the endpoint test - I think you need to update roc locally. No need to comment test. The related fix was released recently https://github.com/roc-streaming/roc-toolkit/releases/tag/v0.2.3

ortex:
still not asserting success context closing for `when there are senders and receivers` case :)

Maybe delete `defer func()` for receiver and sender and add something like that to the end?

```go
	if tt.hasReceivers || tt.hasSenders {
		if tt.hasReceivers {
			err = receiver.Close()
			require.NoError(t, err)
		}
		if tt.hasSenders {
			err = sender.Close()
			require.NoError(t, err)
		}

		err = ctx.Close()
		require.NoError(t, err)
	}
```

ortex:
need to revert https://github.com/roc-streaming/roc-go/pull/58/commits/c1834d21c193d7fbd51bd2c8dcfe32879a0a8cfe everything else LGTM

Darrellbor:
@ortex this block still causes the test to fail both on my branch and on main

ortex:
which block? If you are about endpoint_test you need to update roc itself (locally), not roc-go.

As you can see this test is enabled in main branch and CI is green: https://github.com/roc-streaming/roc-go/actions/runs/4409507925

Darrellbor:
resolved

roc/context_test.go
```diff
@@ -36,3 +36,93 @@ func TestContext_Open(t *testing.T) {
 		})
 	}
 }
+
+func TestContext_Close(t *testing.T) {
+	var (
+		receiver *Receiver
+		sender   *Sender
+	)
```
ortex, line 44: let's move these vars inside `t.Run(tt.name, func(t *testing.T) {` ?

ortex (approved):

gavv:
Cosmetic fix: 005ae7ef1d919f8398147d7f0f53e70aaa46e258

PR #60, opened 2023-03-10
refactor endpoint_test: add subtest name

ortex:
refactor endpoint_test: add subtest name


gavv:
Discussed to add names to all tests.

PR #61, opened 2023-03-13
Add version information to API

### LINKED ISSUE #47: Add version information to API
C library provides version info: https://roc-streaming.org/toolkit/docs/api/reference.html#roc-version

It would be nice to add this feature to the go library too. We need to provide two versions actually: version of the native library (libroc), and version of the bindings (which may be slightly different).

Versions of the bindings can be obtained using https://pkg.go.dev/runtime/debug

dhanusaputra:
close #47

### DIFF
--- a/roc/version.go
+++ b/roc/version.go
@@ -16,7 +16,7 @@ var (
 	versionOnce sync.Once
 )
 
-const bindingsVersion = "0.2.0"
+const bindingsVersion = "0.1.0"
 
 // Version components.
 type Versions struct {
--- a/roc/version.go
+++ b/roc/version.go
@@ -50,22 +50,22 @@ func Version() Versions {
 }
 
 func parseVersion(s string) SemanticVersion {
-	bvs := strings.SplitN(s, ".", 3)
-	if len(bvs) != 3 {
+	vs := strings.SplitN(s, ".", 3)
+	if len(vs) != 3 {
 		panic("semantic version doesn't have 3 parts")
 	}
 
 	v := SemanticVersion{}
 	var err error
-	v.Major, err = strconv.ParseUint(bvs[0], 10, 64)
+	v.Major, err = strconv.ParseUint(vs[0], 10, 64)
 	if err != nil {
 		panic(err)
 	}
-	v.Minor, err = strconv.ParseUint(bvs[1], 10, 64)
+	v.Minor, err = strconv.ParseUint(vs[1], 10, 64)
 	if err != nil {
 		panic(err)
 	}
-	v.Patch, err = strconv.ParseUint(bvs[2], 10, 64)
+	v.Patch, err = strconv.ParseUint(vs[2], 10, 64)
 	if err != nil {
 		panic(err)
 	}
--- a/roc/version.go
+++ b/roc/version.go
@@ -1,0 +1,52 @@
+package roc
+
+/*
+#include <roc/version.h>
+*/
+import "C"
+
+import (
+	"runtime/debug"
+	"sync"
+)
+
+var (
+	version     *Version
+	versionOnce sync.Once
+)
+
+// Version components.
+type Version struct {
+	Library *SemanticVersion // Native library version, librock version.
+	Binding string           // Binding version.
+}
+
+// Semantic version components.
+type SemanticVersion struct {
+	Major uint // Major version component.
+	Minor uint // Minor version component.
+	Patch uint // Patch version component.
+}
+
+// Retrieve version numbers.
+// This function can be used to retrieve actual run-time version of the library.
+// It may be different from the compile-time version when using shared library.
+func GetVersion() *Version {
+	versionOnce.Do(func() {
+		version = &Version{}
+
+		var cVersion C.struct_roc_version
+		C.roc_version_get(&cVersion)
+		version.Library = &SemanticVersion{
+			Major: uint(cVersion.major),
+			Minor: uint(cVersion.minor),
+			Patch: uint(cVersion.patch),
+		}
+
+		if info, ok := debug.ReadBuildInfo(); ok {
+			version.Binding = info.Main.Version
+		}
+	})
+
+	return version
+}
--- a/roc/version.go
+++ b/roc/version.go
@@ -16,7 +16,7 @@ var (
 	versionOnce sync.Once
 )
 
-const bindingsVersion = "0.2.1"
+const bindingsVersion = "0.2.0"
 
 // Version components.
 type Versions struct {
--- a/roc/version.go
+++ b/roc/version.go
@@ -7,46 +7,55 @@ import "C"
 
 import (
 	"runtime/debug"
+	"strconv"
+	"strings"
 	"sync"
 )
 
 var (
-	version     *Version
+	versions    *Versions
 	versionOnce sync.Once
 )
 
 // Version components.
-type Version struct {
-	Library *SemanticVersion // Native library version, librock version.
-	Binding string           // Binding version.
+type Versions struct {
+	Library SemanticVersion // Native library version, libroc version.
+	Binding SemanticVersion // Go binding version.
 }
 
 // Semantic version components.
 type SemanticVersion struct {
-	Major uint // Major version component.
-	Minor uint // Minor version component.
-	Patch uint // Patch version component.
+	Major uint64 // Major version component.
+	Minor uint64 // Minor version component.
+	Patch uint64 // Patch version component.
 }
 
 // Retrieve version numbers.
 // This function can be used to retrieve actual run-time version of the library.
 // It may be different from the compile-time version when using shared library.
-func GetVersion() *Version {
+func Version() Versions {
 	versionOnce.Do(func() {
-		version = &Version{}
+		versions = &Versions{}
 
 		var cVersion C.struct_roc_version
 		C.roc_version_get(&cVersion)
-		version.Library = &SemanticVersion{
-			Major: uint(cVersion.major),
-			Minor: uint(cVersion.minor),
-			Patch: uint(cVersion.patch),
+		versions.Library = SemanticVersion{
+			Major: uint64(cVersion.major),
+			Minor: uint64(cVersion.minor),
+			Patch: uint64(cVersion.patch),
 		}
 
 		if info, ok := debug.ReadBuildInfo(); ok {
-			version.Binding = info.Main.Version
+			v := strings.TrimPrefix(info.Main.Version, "v")
+			vs := strings.SplitN(v, ".", 3)
+			if len(vs) != 3 {
+				return
+			}
+			versions.Binding.Major, _ = strconv.ParseUint(vs[0], 10, 64)
+			versions.Binding.Minor, _ = strconv.ParseUint(vs[1], 10, 64)
+			versions.Binding.Patch, _ = strconv.ParseUint(vs[2], 10, 64)
 		}
 	})
 
-	return version
+	return *versions
 }
--- a/roc/version.go
+++ b/roc/version.go
@@ -6,16 +6,17 @@ package roc
 import "C"
 
 import (
-	"runtime/debug"
 	"strconv"
 	"strings"
 	"sync"
 )
 
 var (
-	versions    *Versions
+	versions    Versions
 	versionOnce sync.Once
 )
+
+const bindingsVersion = "0.2.0"
 
 // Version components.
 type Versions struct {
@@ -35,8 +36,6 @@ type SemanticVersion struct {
 // It may be different from the compile-time version when using shared library.
 func Version() Versions {
 	versionOnce.Do(func() {
-		versions = &Versions{}
-
 		var cVersion C.struct_roc_version
 		C.roc_version_get(&cVersion)
 		versions.Library = SemanticVersion{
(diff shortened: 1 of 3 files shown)


roc/version.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+var (
+	version     *Version
+	versionOnce sync.Once
+)
+
+// Version components.
+type Version struct {
+	Library *SemanticVersion // Native library version, librock version.
```
gavv, line 20: nit:

* we can remove pointer
* librock => libroc :)

roc/version.go
```diff
@@ -0,0 +1,52 @@
...
+	Binding string           // Binding version.
```
gavv, line 21: * Can we parse string into components and use SemanticVersion for this field too?
* Binding => Go bindings

roc/version.go
```diff
@@ -0,0 +1,52 @@
...
+type SemanticVersion struct {
+	Major uint // Major version component.
+	Minor uint // Minor version component.
+	Patch uint // Patch version component.
+}
+
+// Retrieve version numbers.
+// This function can be used to retrieve actual run-time version of the library.
+// It may be different from the compile-time version when using shared library.
+func GetVersion() *Version {
```
gavv, line 34: nit:

* we can remove the pointer to make it clear that it's never nil
* maybe rename it from GetVersion to just Version for simplicity? roc.Version() looks better IMHO
* maybe we can also rename struct Version to struct Versions to avoid confusion with the function

roc/version_test.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestVersion_Get(t *testing.T) {
+	v := GetVersion()
+	require.NotEmpty(t, v)
+	require.NotEmpty(t, v.Library)
```
gavv, line 12: We can also test that:

* at least one of the version components is non-zero
* major versions of C and Go are the same
* minor version of Go >= minor version of C

These invariants should be always true.

gavv (commented):
Thanks for PR! A few minor comments

roc/version_test.go
dhanusaputra: for Go version seems not able to get it during unit testing
only available after it's built
so don't check it

dhanusaputra: btw not sure how to build this, there is no main.go

gavv:
BTW welcome to matrix chat! https://roc-streaming.org/toolkit/docs/about_project/contacts.html

roc/version.go
```diff
@@ -0,0 +1,61 @@
...
+
+		var cVersion C.struct_roc_version
+		C.roc_version_get(&cVersion)
+		versions.Library = SemanticVersion{
+			Major: uint64(cVersion.major),
+			Minor: uint64(cVersion.minor),
+			Patch: uint64(cVersion.patch),
+		}
+
+		if info, ok := debug.ReadBuildInfo(); ok {
```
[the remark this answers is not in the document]
dhanusaputra: instead of notify us via CI, maybe can try push the newer version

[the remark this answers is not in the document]
dhanusaputra: seems need to give permission to github actions first before it can push some changes
so updated to check only

roc/version.go
```diff
@@ -0,0 +1,57 @@
...
+	Patch uint64 // Patch version component.
+}
+
+// Retrieve version numbers.
+// This function can be used to retrieve actual run-time version of the library.
+// It may be different from the compile-time version when using shared library.
+func Version() Versions {
+	versionOnce.Do(func() {
+		var cVersion C.struct_roc_version
+		C.roc_version_get(&cVersion)
```
Asalle, line 40: I think it's generally beneficial to check what do functions return, here for example some weird stuff can happen and C.roc_version_get can fail, then we'll just panic with no explanation. A more graceful shutdown stating that roc_version_get crapped out would be preferable.

Asalle:
Amazing job! It would make it even better if you included some motivation / `closes #XX` in the commit message to document it a bit more.

roc/version_test.go
```diff
@@ -0,0 +1,16 @@
+package roc
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestVersion(t *testing.T) {
```
Asalle, line 9: Great test, although it does only test a successful path. I can see that the way `Version` func is implemented you cannot really inject any test strings into it and test failures here. Have you thought about decoupling it to make it possible? There are a lot of moving parts in `Version` and it can crap out in e. g. lines 40, 47, 51, 52 and 53. I am not sure if it is vital, but I would suggest it.

roc/version_test.go
Asalle: weird, the `roc_get_version` should be definitely available during unit tests, just to smoke test, do you have libroc installed? I wrote a simple sanity test to check it:
```
package main

/*
#cgo LDFLAGS: -lroc

#include <roc/version.h>
*/
import "C"

import (
        "fmt"
)

func main() {
        var cVersion C.struct_roc_version
        C.roc_get_version(&cVersion)
        fmt.Print(cVersion)
}
```
My output is 
```
 go run /tmp/test.go
{0 2 1}%    
```

I think I have an old version of roc where the method for getting the version is named differently, please rename it if it does not work. Let me know if you have any problems with this!

Asalle: > btw not sure how to build this, there is no main.go

there's no main.go because it's a library, basically just a set of functions that can be used by applications that do have main.go. Unit tests pull on those methods (by having a func main declared somewhere under the hood), so if you want to test something, write a test for it with go unit testing framework or make a small program like I did, but don't commit it unless it's in the unit test form.

dhanusaputra: > I think I have an old version of roc where the method for getting the version is named differently, please rename it if it does not work. Let me know if you have any problems with this!

yes in newer version it's renamed https://roc-streaming.org/toolkit/docs/development/changelog.html#version-0-2-2-feb-27-2023

> there's no main.go because it's a library, basically just a set of functions that can be used by applications that do have main.go. Unit tests pull on those methods (by having a func main declared somewhere under the hood), so if you want to test something, write a test for it with go unit testing framework or make a small program like I did, but don't commit it unless it's in the unit test form.

got it, thanks for your explanation!

roc/version.go
dhanusaputra: yup agree but seems from [API doc](https://roc-streaming.org/toolkit/docs/api/reference.html#roc-version) the roc_version_get will not return error, please CMIIW thanks

roc/version_test.go
dhanusaputra: > Have you thought about decoupling it to make it possible? 

yes already thought about this, but seems ok to me to ignore and try to return as much as possible instead of returning error

> There are a lot of moving parts in Version and it can crap out in e. g. lines 40, 47, 51, 52 and 53. I am not sure if it is vital, but I would suggest it.

yes those lines can fail when parsing invalid `bindingsVersion`
let me try add 1 unit test to check whether the `bindingsVersion` is valid or not so we can check it early in unit test

roc/version.go
Asalle: I see, then I guess the API designers agreed on some null-like version (e. g. 0.0.0) that will indicate an error? @gavv would appreciate your input

roc/version_test.go
Asalle: > yes already thought about this, but seems ok to me to ignore and try to return as much as possible instead of returning error

In general, I would not think it's ok to do it, but I can see your motivation, because `Version` returns {0,0,0} in case of error anyway. Let's see how that discussion plays out, so depending on what we decide, you would leave this func as is or rewrite it a bit to decouple getting the version from C and parsing it to allow for more modularity and testability.

roc/version_test.go
```diff
@@ -0,0 +1,29 @@
...
+	v := Version()
+	require.NotZero(t, v.Library.Major+v.Library.Minor+v.Library.Patch)
+	require.NotZero(t, v.Bindings.Major+v.Bindings.Minor+v.Bindings.Patch)
+	require.Equal(t, v.Library.Major, v.Bindings.Major)
+	require.GreaterOrEqual(t, v.Bindings.Minor, v.Library.Minor)
+}
+
+func Test_bindingsVersion(t *testing.T) {
+	bvs := strings.SplitN(bindingsVersion, ".", 3)
```
Asalle, line 20: Thanks for the update! Does it pass locally for you? I'm not sure where does the var `bindingsVersion` come from, and on the CI it this test seems to be skipped. I haven't run it locally, let me try it later.

roc/version_test.go
Asalle: did that small test of mine work for you? if not - let me know, I'd be happy to help :rocket:

dhanusaputra: yes the test works

roc/version_test.go
dhanusaputra: > Does it pass locally for you?

yes

> I'm not sure where does the var bindingsVersion come from, and on the CI it this test seems to be skipped

it's from [latest release](https://github.com/roc-streaming/roc-go/releases), it's skipped because need to push changes with tag

roc/version_test.go
gavv: @Asalle 

> weird, the roc_get_version should be definitely available during unit tests

I think what was missing is **bindings** version (which we tried to get using ReadBuildInfo), not libroc version. See https://github.com/roc-streaming/roc-go/pull/61#discussion_r1145008792

roc/version.go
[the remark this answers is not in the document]
gavv: > instead of notify us via CI, maybe can try push the newer version

What do you mean by pushing new version?

roc/version.go
gavv: @Asalle it just can't fail, because it returns version hard-coded during build; thus no error code

roc/version_test.go
gavv: >  let me try add 1 unit test to check whether the bindingsVersion is valid or not so we can check it early in unit test

I like this idea. User interface is simpler (no need to check for errors from roc.Version), and at the same time our test suite guarantees that failures are not possible (by validating hard-coded version).

The only downside is that the test duplicates parsing logic. If we update version.go and forget version_test.go, then the test's validation may become irrelevant.

So I'd suggest to refactor it to reuse parsing code. One way to do it:

1. extract private function parseVersion that accepts string, returns version triple, or panics on invalid string format
2. in roc.Version, use parseVersion
3. in tests, check that:
  - parseVersion works on valid input
  - parseVersion panics on invalid input
  - parseVersion works in hard-coded bindingsVersion

roc/version.go
```diff
@@ -0,0 +1,57 @@
...
+var (
+	versions    Versions
+	versionOnce sync.Once
+)
+
+const bindingsVersion = "0.2.0"
+
+// Version components.
+type Versions struct {
+	Library  SemanticVersion // Native library version, libroc version.
```
gavv, line 23: I think it will be more clear if we rename "Library" to "Native"

roc/version_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+func TestVersion(t *testing.T) {
+	v := Version()
+	require.NotZero(t, v.Native.Major+v.Native.Minor+v.Native.Patch)
+	require.NotZero(t, v.Bindings.Major+v.Bindings.Minor+v.Bindings.Patch)
+	require.Equal(t, v.Native.Major, v.Bindings.Major)
+	require.GreaterOrEqual(t, v.Bindings.Minor, v.Native.Minor)
+}
+
+func Test_parseVersion(t *testing.T) {
```
gavv, line 17: nit: I suggest to rename tests to something like this:

* TestVersion => TestVersion_Get
* Test_parseVersion => TestVersion_Parse

For consistency with other tests, where we name them `Test<File>_<Component>`.

roc/version_test.go
```diff
@@ -0,0 +1,86 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.wantPanic {
+				require.Panics(t, func() { parseVersion(tt.args.s) })
+				return
+			}
+			var v SemanticVersion
+			require.NotPanics(t, func() { v = parseVersion(tt.args.s) })
+			if tt.want != (SemanticVersion{}) {
+				require.Equal(t, tt.want, v)
+			}
```
gavv, line 83: nit: we can simplify this a bit:

```go
			if tt.wantPanic {
				require.Panics(t, func() { parseVersion(tt.args.s) })
			} else {
				require.Equal(t, tt.want, parseVersion(tt.args.s))
			}
```

(we can require that either wantPanic is true of want is set; we don't need NotPanics I think)

roc/version_test.go
gavv: +1 from me, @Asalle feel free to resolve this

gavv:
Added 4 minor comments, otherwise looks fine. Waiting for review from Asa.

gavv (approved):
LGTM

roc/version_test.go
Asalle: Oh I see, had to get some context to review this PR properly

roc/version.go
Asalle: oh, I see, got it, then the implementation does makes sense. Good job @dhanusaputra!

roc/version_test.go
Asalle: I remember injecting an env var during build [like so, with ldflags](https://stackoverflow.com/questions/68051479/how-to-go-build-ldflags-with-custom-environment-variable), didn't try it with libs though, should work with go build, right? Worth a try! Seems like [you can inject that stuff into packages as well, not just main](https://goenning.net/2017/01/25/adding-custom-data-go-binaries-compile-time/).

Asalle (approved):
Again, amazing job and thanks for your contribution!

roc/version_test.go
gavv: Discussion moved to chat

Issue #41, opened 2022-12-23
Add unit tests for sender and receiver errors


gavv:
It would be nice to cover with unit tests errors returned by:

* OpenSender, OpenReceiver
* Sender.Connect, Receiver.Bind
* Sender.WriteFloats, Receiver.ReadFloats

See:

https://coveralls.io/builds/55438429/source?filename=roc%2Fsender.go
https://coveralls.io/builds/55438429/source?filename=roc%2Freceiver.go

No need to cover panics: mostly there is no way to cover them.

Darrellbor:
@gavv I would like to handle this

gavv:
Landed

PR #65, opened 2023-03-25
[41]:... unit tests for sender and receiver errors

Darrellbor:
PR for #41 
Implemented 
- [x] `TestReceiver_Bind`
- [x] `TestReceiver_ReadFloats`
- [x] `TestSender_Connect`
- [x] `TestSender_WriteFloats`

### DIFF
--- a/roc/sender_test.go
+++ b/roc/sender_test.go
@@ -136,7 +136,7 @@ func TestSender_Connect(t *testing.T) {
 		slot               Slot
 		iface              Interface
 		senderClosedBefore bool
-		endpoint             *Endpoint
+		endpoint           *Endpoint
 		wantErr            error
 	}{
 		{
@@ -144,7 +144,7 @@ func TestSender_Connect(t *testing.T) {
 			slot:               SlotDefault,
 			iface:              InterfaceAudioSource,
 			senderClosedBefore: false,
-			endpoint:             baseEndpoint,
+			endpoint:           baseEndpoint,
 			wantErr:            nil,
 		},
 		{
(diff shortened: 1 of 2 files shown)


Darrellbor:
@gavv @ortex PR ready for review

gavv:
Also @Asalle

ortex:
Nice work!

a couple of tests could be added:

- for Sender.Connect, Receiver.Bind: add tests with an invalid endpoint that fails to convert to `C endpoint`, e.g. with invalid protocol `Protocol(1)`
- for Sender.Writefloat, Receiver.ReadFloats: add tests with zero-sized frame `[]float32{}`

also, point for discussion (@gavv ):

Maybe add test `Receiver_Closed` (or something like that) and remove `receiverClosedBefore` scenario from all tests?
Because we need to check that each function call (except `Сlose()`) on a closed receiver return an error 
e.g:
```go

func TestReceiver_Closed(t *testing.T) {
	tests := []struct {
		name      string
		operation func(receiver *Receiver) error
	}{
		{
			name: "SetMulticastGroup after close",
			operation: func(receiver *Receiver) error {
				return receiver.SetMulticastGroup(SlotDefault, InterfaceAudioSource, "127.0.0.1")
			},
		},
		{
			name: "Bind after close",
			operation: func(receiver *Receiver) error {
				return receiver.Bind(SlotDefault, InterfaceAudioSource, nil)
			},
		},
		{
			name: "ReadFloats after close",
			operation: func(receiver *Receiver) error {
				recFloats := make([]float32, 2)
				return receiver.ReadFloats(recFloats)
			},
		},
	}
	for _, tt := range tests {
		ctx, err := OpenContext(ContextConfig{})
		require.NoError(t, err)
		require.NotNil(t, ctx)

		receiver, err := OpenReceiver(ctx, makeReceiverConfig())
		require.NoError(t, err)
		require.NotNil(t, receiver)

		err = receiver.Close()
		require.NoError(t, err)

		require.Equal(t, errors.New("receiver is closed"), tt.operation(receiver))

		err = ctx.Close()
		require.NoError(t, err)
	}
}
```

gavv:
I agree with Andrey, separating this check will make the table tests more uniform, which is the point of tables. Ideally table tests should have minimal amount of conditions and if tests have rather different flow, they should belong to different tables / tests.

Darrellbor:
@gavv @ortex 
reviewed based on the comments and ready for re-review

roc/sender_test.go
```diff
@@ -103,25 +92,194 @@ func TestSender_SetReuseaddr(t *testing.T) {
...
+			err = sender.Close()
+			require.NoError(t, err)
+
 			err = ctx.Close()
 			require.NoError(t, err)
 		})
 	}
 }
+
+func TestSender_Close(t *testing.T) {
```
ortex, line 243: can we add test for `SetReuseaddr` as well?

roc/sender_test.go
```diff
@@ -103,25 +92,194 @@ func TestSender_SetReuseaddr(t *testing.T) {
...
+			wantErr:  newNativeErr("roc_endpoint_set_protocol()", -1),
+		},
+	}
+
+	for _, tt := range cases {
+		t.Run(tt.name, func(t *testing.T) {
+			ctx, err := OpenContext(ContextConfig{})
+			require.NoError(t, err)
+
+			sender, err := OpenSender(ctx, SenderConfig{
```
ortex, line 157: we could use `makeSenderConfig` here as in the receiver test

ortex (approved):

Issue #67, opened 2023-03-28
Improve receiver and sender test coverage


gavv:
See coverage here:

* https://coveralls.io/builds/58298808/source?filename=roc%2Freceiver.go
* https://coveralls.io/builds/58298808/source?filename=roc%2Fsender.go

Cover the following branches:

* Receiver:

   * OpenReceiver: nil context
   * OpenReceiver: closed context

* Sender:

   * OpenSender: nil context
   * OpenSender: closed context

gavv:
See also #43, #44.

PR #68, opened 2023-03-28
add tests for sender, receiver open errors

ortex:
#67

### DIFF
--- a/roc/receiver_test.go
+++ b/roc/receiver_test.go
@@ -17,20 +17,18 @@ func TestReceiver_Open(t *testing.T) {
 		{
 			name: "ok",
 			contextFunc: func() *Context {
-				ctx, _ := OpenContext(ContextConfig{})
+				ctx, err := OpenContext(ContextConfig{})
+				require.NoError(t, err)
 				return ctx
 			},
-			config: ReceiverConfig{
-				FrameSampleRate: 44100,
-				FrameChannels:   ChannelSetStereo,
-				FrameEncoding:   FrameEncodingPcmFloat,
-			},
+			config:  makeReceiverConfig(),
 			wantErr: nil,
 		},
 		{
 			name: "invalid config",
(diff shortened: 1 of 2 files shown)


gavv:
We could nicely merge new tests into TestXxx_Open by adding context creation func to the table:

```go
		{
			name:    "closed context",
			...
			contextFunc: func() *Context {
				ctx, err := OpenContext()
				ctx.Close()
				return ctx
			},
		},
```

WDYT?

roc/receiver_test.go
```diff
@@ -9,12 +9,17 @@ import (
...
-		wantErr error
+		name        string
+		contextFunc func() *Context
+		config      ReceiverConfig
+		wantErr     error
 	}{
 		{
 			name: "ok",
+			contextFunc: func() *Context {
+				ctx, _ := OpenContext(ContextConfig{})
```
gavv, line 20: I'd suggest to check that error is nil here and in other context funcs.

roc/receiver_test.go
```diff
@@ -9,12 +9,17 @@ import (
...
+				return ctx
+			},
 			config: ReceiverConfig{
 				FrameSampleRate: 44100,
 				FrameChannels:   ChannelSetStereo,
```
gavv, line 25: Let's use makeReceiverConfig here?

roc/receiver_test.go
gavv: Same for sender

roc/receiver_test.go
gavv: Same for sender

roc/receiver_test.go
ortex: done

roc/receiver_test.go
ortex: agreed. Done

gavv:
LGTM!

PR #74, opened 2023-04-08
Logger improvements

gavv:
This PR contains two commits:

* minor refactoring in log.go
* decoupling of C and Go logging threads

More details about the second commit:

I noticed that when user calls SetLogger/SetLoggerFunc to provide custom log handler, or configures standard logger to use custom output writer, it may happen that our C threads invoke arbitrary complex user's go code.

I feel that invoking random user's go code on random C threads may be dangerous, at best it can confuse go scheduler and probably cause performance issues, though I don't have specific examples on top of my head.

Anyway, I just decoupled this threads: C callback now writes messages to a channel, and user's go code is invoked in a separate goroutine which reads messages from that channel. Now we can be sure that user's callback stay in "go world" and callbacks invoked on C thread is small and minimal and is completely controlled by us.

### DIFF
--- a/roc/log.go
+++ b/roc/log.go
@@ -84,25 +84,28 @@ func rocGoLogHandler(cMessage *C.roc_log_message) {
 	loggerMu.Lock()
 	defer loggerMu.Unlock()
 
-	if loggerFunc != nil {
-		message := LogMessage{
-			Level: LogLevel(cMessage.level),
-			Time:  uint64(cMessage.time),
-			Pid:   uint64(cMessage.pid),
-			Tid:   uint64(cMessage.tid),
-		}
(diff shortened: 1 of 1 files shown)


ortex:
LGTM

slightly related stuff:
Need to fix comments for LogInfo and LogTrace levels.

```go
	// LogError enables informational messages and above.
	LogInfo LogLevel = 2
...
	// LogDebug enables extra verbose logging, which may hurt performance
	// and should not be used in production.
	LogTrace LogLevel = 4
```

gavv:
> Need to fix comments for LogInfo and LogTrace levels.

Fixed.

ortex (approved):

Issue #71, opened 2023-04-03
Add version checking


gavv:
When any non-method public function is called first time (except Version func), we need to call Version() to retrieve versions of native lib and go bindings, and check the following invariant:

* major versions are equal
* minor version of bindings is greater than or equal to minor version of native lib

(See "Versioning" section in Readme for explanation)

Non-method public functions include: OpenContext, OpenSender, OpenReceicer, SetLogger, SetLoggerFunc, SetLogLevel.

If the invariant is not satisfied, we should panic. Panic message should include both versions and reason.

fastbyt3:
Hey o/ Can I take a shot on this?

gavv:
Sure, thanks!

fastbyt3:
Sorry but I'm pretty new to open source contribs. Is it cool with you if i DM (any platform works) you regarding some qs i have?

gavv:
@fastbyt3 welcome to join `#roc-streaming-dev:matrix.org` on matrix

https://roc-streaming.org/toolkit/docs/about_project/contacts.html

PR #75, opened 2023-04-09
implemented version checking

fastbyt3:
Attempted to complete #71. Please check and let me know if i have to change the logic

### DIFF
--- a/roc/version.go
+++ b/roc/version.go
@@ -6,6 +6,7 @@ package roc
 import "C"
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
@@ -78,31 +79,35 @@ func stringifyVersion(sv SemanticVersion) string {
 	return fmt.Sprintf("%d.%d.%d", sv.Major, sv.Minor, sv.Patch)
 }
 
+func checkVersionCompatibility(nativeVersion, bindingsVersion SemanticVersion) error {
+	if nativeVersion.Major != bindingsVersion.Major {
+		return errors.New(fmt.Sprintf(`
+Bindings are compatible with native C library only if Major versions are same
+Bindings version: %s
+C library version: %s`,
+			stringifyVersion(bindingsVersion), stringifyVersion(nativeVersion),
+		))
+	}
+
+	if nativeVersion.Minor > bindingsVersion.Minor {
+		return errors.New(fmt.Sprintf(`
+Bindings are compatible with native C library only if its Minor version is same or higher
+Bindings version: %s
+C library version: %s`,
+			stringifyVersion(bindingsVersion), stringifyVersion(nativeVersion),
+		))
+	}
+	return nil
+}
+
 var runVersionCheckOnce sync.Once
 
 func versionCheck() {
 	runVersionCheckOnce.Do(func() {
 		v := Version()
-		nativeVersion := v.Native
-		bindingsVersion := v.Bindings
-
-		if nativeVersion.Major != bindingsVersion.Major {
-			panic(fmt.Sprintf(`
-Bindings are compatible with native C library only if Major versions are same
-Bindings version: %s
-C library version: %s`,
-				stringifyVersion(bindingsVersion), stringifyVersion(nativeVersion),
-			))
+		err := checkVersionCompatibility(v.Native, v.Bindings)
+		if err != nil {
+			panic(err.Error())
 		}
-
-		if nativeVersion.Minor > bindingsVersion.Minor {
-			panic(fmt.Sprintf(`
-Bindings are compatible with native C library only if its Minor version is same or higher
-Bindings version: %s
-C library version: %s`,
-				stringifyVersion(bindingsVersion), stringifyVersion(nativeVersion),
-			))
-		}
-		return
 	})
 }
--- a/roc/version.go
+++ b/roc/version.go
@@ -81,26 +81,28 @@ func stringifyVersion(sv SemanticVersion) string {
 var runVersionCheckOnce sync.Once
 
 func versionCheck() {
-	v := Version()
-	nativeVersion := v.Native
-	bindingsVersion := v.Bindings
+	runVersionCheckOnce.Do(func() {
+		v := Version()
+		nativeVersion := v.Native
+		bindingsVersion := v.Bindings
 
-	if nativeVersion.Major != bindingsVersion.Major {
-		panic(fmt.Sprintf(`
+		if nativeVersion.Major != bindingsVersion.Major {
+			panic(fmt.Sprintf(`
 Bindings are compatible with native C library only if Major versions are same
 Bindings version: %s
 C library version: %s`,
-			stringifyVersion(bindingsVersion), stringifyVersion(nativeVersion),
-		))
-	}
+				stringifyVersion(bindingsVersion), stringifyVersion(nativeVersion),
+			))
+		}
 
-	if nativeVersion.Minor > bindingsVersion.Minor {
-		panic(fmt.Sprintf(`
+		if nativeVersion.Minor > bindingsVersion.Minor {
+			panic(fmt.Sprintf(`
 Bindings are compatible with native C library only if its Minor version is same or higher
 Bindings version: %s
 C library version: %s`,
-			stringifyVersion(bindingsVersion), stringifyVersion(nativeVersion),
-		))
-	}
-	return
+				stringifyVersion(bindingsVersion), stringifyVersion(nativeVersion),
+			))
+		}
+		return
+	})
 }
--- a/roc/version.go
+++ b/roc/version.go
@@ -6,6 +6,7 @@ package roc
 import "C"
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 	"sync"
@@ -71,4 +72,35 @@ func parseVersion(s string) SemanticVersion {
 	}
 
 	return v
+}
+
+func stringifyVersion(sv SemanticVersion) string {
+	return fmt.Sprintf("%d.%d.%d", sv.Major, sv.Minor, sv.Patch)
+}
+
+var runVersionCheckOnce sync.Once
+
+func versionCheck() {
+	v := Version()
+	nativeVersion := v.Native
+	bindingsVersion := v.Bindings
+
+	if nativeVersion.Major != bindingsVersion.Major {
+		panic(fmt.Sprintf(`
+Bindings are compatible with native C library only if Major versions are same
+Bindings version: %s
+C library version: %s`,
+			stringifyVersion(bindingsVersion), stringifyVersion(nativeVersion),
+		))
+	}
+
+	if nativeVersion.Minor > bindingsVersion.Minor {
+		panic(fmt.Sprintf(`
+Bindings are compatible with native C library only if its Minor version is same or higher
+Bindings version: %s
+C library version: %s`,
+			stringifyVersion(bindingsVersion), stringifyVersion(nativeVersion),
+		))
+	}
+	return
 }
--- a/roc/version.go
+++ b/roc/version.go
@@ -14,17 +14,14 @@ import (
 )
 
 var (
-	versions    Versions
-	versionOnce sync.Once
+	versionInfo      VersionInfo
+	versionOnce      sync.Once
+	versionCheckOnce int32
 )
 
-const bindingsVersion = "0.2.0"
+var versionCheckFn = versionCheck
 
-// Version components.
-type Versions struct {
-	Native   SemanticVersion // Native library version, libroc version.
-	Bindings SemanticVersion // Go bindings version.
-}
+const bindingsVersion = "0.2.0"
 
 // Semantic version components.
 type SemanticVersion struct {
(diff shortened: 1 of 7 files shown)


gavv:
Thanks for PR!

Two preliminary comments:

* please move runVersionCheckOnce.Do inside versionCheck, to avoid copying it each time
* please add a test for versionCheck

How you could add a test:

* extract a function, say, checkVersionCompatibility, that accepts two versions, do the checks, and panic if needed
* cover it with tests, passing to it different versions
* from your versionCheck function, call checkVersionCompatibility with actual versions

roc/version.go
```diff
@@ -72,3 +73,40 @@ func parseVersion(s string) SemanticVersion {
 
 	return v
 }
+
+func stringifyVersion(sv SemanticVersion) string {
+	return fmt.Sprintf("%d.%d.%d", sv.Major, sv.Minor, sv.Patch)
+}
```
gavv, line 79: nit: instead of stringifyVersion, we can add SemanticVersion.String() method, I think it would be nicer, and versions will automatically support `%s` formatting both for us and for users.

roc/version.go
```diff
@@ -72,3 +73,40 @@ func parseVersion(s string) SemanticVersion {
...
+
+func checkVersionCompatibility(nativeVersion, bindingsVersion SemanticVersion) error {
```
gavv, line 81: nit: we can replace two args with one arg `Versions`. Also maybe rename to just `checkCompatibility`?

I think `checkCompatibility(versions)` is not ambiguous and reads well.

roc/version.go
```diff
@@ -72,3 +73,40 @@ func parseVersion(s string) SemanticVersion {
...
+Bindings are compatible with native C library only if its Minor version is same or higher
+Bindings version: %s
+C library version: %s`,
+			stringifyVersion(bindingsVersion), stringifyVersion(nativeVersion),
+		)
+	}
+	return nil
+}
+
+var runVersionCheckOnce sync.Once
```
gavv, line 102: nit: maybe rename to versionCheckOnce or versionCheckGuard? For clarify and for easier association with versionCheck() func.

roc/version_test.go
```diff
@@ -80,3 +80,45 @@ func TestVersion_Parse(t *testing.T) {
...
+
+func TestCheck_Version_Compatibility(t *testing.T) {
+	tests := []struct {
+		name            string
+		nativeVersion   SemanticVersion
+		bindingsVersion SemanticVersion
+		wantErr         bool
+	}{
+		{
+			name:            "Compatible versions",
```
gavv, line 92: nit: in other tests, subtest names are lowercase

gavv (commented):
Nice, thanks for update! A few more comments.

roc/version.go
```diff
@@ -72,3 +73,40 @@ func parseVersion(s string) SemanticVersion {
...
+func stringifyVersion(sv SemanticVersion) string {
+	return fmt.Sprintf("%d.%d.%d", sv.Major, sv.Minor, sv.Patch)
+}
+
+func checkVersionCompatibility(nativeVersion, bindingsVersion SemanticVersion) error {
+	if nativeVersion.Major != bindingsVersion.Major {
+		return fmt.Errorf(`
+Bindings are compatible with native C library only if Major versions are same
+Bindings version: %s
+C library version: %s`,
```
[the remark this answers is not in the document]
fastbyt3:
```
detected incompatibility between roc bindings ( 1.1.1 ) and native library ( 1.2.5): minor version of binding less than native
```
Is this format okay?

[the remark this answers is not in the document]
gavv: perfect

gavv:
Given it another thought..

Probably some users may want to handle versions failure gracefully instead of crashing. For those users, we could provide a public api to make an explicit version check before using the library. 

One way todo it is to convert checkVersionCompatibility to a public method of Versions struct, say, Versions.Validate, that returns error.

If user wants, they can call it and handle error.

The rest remains the same, in entry points we call versionCheck, which calls Versions.Valudate and panics on error.

In this case versionCheck can be probably renamed to something like validateVersions.

What do you think?

gavv:
Also we forgot another entry point - ParseEndpoint.

gavv:
Also, could you please add a comment for versionCheck func, stating that it must be called from all lib entry points and explaining that entry points are exported non-method funcs?

fastbyt3:
> One way to do it is to convert checkVersionCompatibility to a public method of Versions struct, say, Versions.Validate, that returns error.
> 
> If user wants, they can call it and handle error.
> 
> What do you think?

Yeah this sounds great

roc/version_test.go
```diff
@@ -80,3 +80,45 @@ func TestVersion_Parse(t *testing.T) {
...
+			nativeVersion:   parseVersion("1.1.1"),
+			bindingsVersion: parseVersion("1.0.1"),
+			wantErr:         true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := checkVersionCompatibility(tt.nativeVersion, tt.bindingsVersion)
+			if tt.wantErr {
```
Asalle, line 113: I'm not sure if it's a good idea or not, but as a general information, if you want to test if versionCheck was really called in the API functions like OpenContext, you can use [gomock](https://github.com/golang/mock) for that.

Asalle:
good job @fastbyt3! Please make sure that the checkVersion() call is called in all the API calls it needs to be called from.

roc/version.go
gavv, line 1: I suggest to reorder definitions a bit:

* first place `SemanticVersion` and its methods (SemanticVersion.String())
* then place `Versions` and its methods (Version() and Versions.Validate())
* then place private helpers (parseVersion, versionCheck)
* move global vars versionCheckOnce and versionCheckFn to the top where we already have some other global vars

Also I think we should rename `Versions` to something more distinguishable from `Version`; maybe `VersionInfo`?

roc/version.go
```diff
@@ -72,3 +79,43 @@ func parseVersion(s string) SemanticVersion {
...
+	}
+
+	return nil
+}
+
+var versionCheckOnce int32
+
+// Runs check for version compatibility
+func versionCheck() {
+	if atomic.CompareAndSwapInt32(&versionCheckOnce, 1, 0) {
```
gavv, line 112: It should be `CompareAndSwapInt32(&versionCheckOnce, 0, 1)` because initial value of versionCheckOnce is `0`.

roc/version.go
```diff
@@ -72,3 +79,43 @@ func parseVersion(s string) SemanticVersion {
 
 	return v
 }
+
+// Validate version compatibility of roc bindings and native library.
+// This function must be called at all library entry point atleast once.
+// Entry points refer to exported non-method functions of this package
+// This includes OpenContext, OpenSender, OpenReceiver, ParseEndpoint,
+// SetLogger, SetLoggerFunc and SetLogLevel
+func (version Versions) Validate() error {
```
gavv, line 88: I think this part:

```
// This function must be called at all library entry point atleast once.
// Entry points refer to exported non-method functions of this package
// This includes OpenContext, OpenSender, OpenReceiver, ParseEndpoint,
// SetLogger, SetLoggerFunc and SetLogLevel
```

Should be moved to versionCheck.

Here, in Versions.Validate, comment is for the user, and user is not responsible for calling Validate in entrypoints and does not have to know about it.

In Versions.Validate comment it's better to say something like: "if versions are incompatible, will return error describing the problem, otherwise will return nil".

Also I think we should not list all entrypoints here. We already have one list in TestVersion_Entrypoints. We can easily forget to update list in comment and it will become outdated. 

Also "atleast" => "at least".

roc/version.go
```diff
@@ -72,3 +79,43 @@ func parseVersion(s string) SemanticVersion {
...
+	bindingsVersion := version.Bindings
+	nativeVersion := version.Native
+
+	errMsg := fmt.Sprintf(
+		"Detected incompatibility between roc bindings ( %s ) and native library ( %s ): ",
+		bindingsVersion.String(), nativeVersion.String(),
```
gavv, line 94: nit: no need to explicitly call String().

roc/version.go
```diff
@@ -72,3 +79,43 @@ func parseVersion(s string) SemanticVersion {
...
+	if atomic.CompareAndSwapInt32(&versionCheckOnce, 1, 0) {
+		version := Version()
+		err := version.Validate()
+		if err != nil {
+			panic(err.Error())
+		}
+	}
+}
+
+var versionCheckFn = func() { versionCheck() }
```
gavv, line 121: nit: can be simplified to:

```
var versionCheckFn = versionCheck
```

roc/version_test.go
```diff
@@ -80,3 +81,199 @@ func TestVersion_Parse(t *testing.T) {
...
+				Native: SemanticVersion{
+					Major: 1,
+					Minor: 1,
+					Patch: 1,
+				},
+				Bindings: SemanticVersion{
+					Major: 1,
+					Minor: 1,
+					Patch: 1,
+				},
```
gavv, line 103: nit: maybe tests will be more compact & readable if we omit fields for SemanticVersion? Like:

```go
				Native: SemanticVersion{1, 1, 1},
				Bindings: SemanticVersion{1, 1, 1},
```

roc/version_test.go
```diff
@@ -80,3 +81,199 @@ func TestVersion_Parse(t *testing.T) {
...
+			},
+			wantErr: errors.New(
+				"Detected incompatibility between roc bindings ( 0.1.1 ) and native library ( 1.1.1 ):" +
+					" Major versions are different",
+			),
+		},
+		{
+			name: "incompatible: binding's major greater than native",
```
gavv, line 127: nit: major => major version

roc/version_test.go
```diff
@@ -80,3 +81,199 @@ func TestVersion_Parse(t *testing.T) {
...
+func TestVersion_Check(t *testing.T) {
+	require.NotPanics(t, func() { versionCheck() })
+}
+
+var versionCheckCalled = false
+
+func TestVersion_Entrypoints(t *testing.T) {
+	versionCheckFn = func() {
+		versionCheckCalled = true
+	}
```
gavv, line 217: Better make versionCheckCalled a local variable of subtest, and overwrite versionCheckFn temporary in subtest. Also we should restore original versionCheckFn at the end of the test (using defer, so that it will be restored even on panic).

roc/version_test.go
```diff
@@ -80,3 +81,199 @@ func TestVersion_Parse(t *testing.T) {
...
+
+	tests := []struct {
+		name       string
+		entrypoint func()
+	}{
+		{
+			name: "OpenContext",
+			entrypoint: func() {
+				ctx, _ := OpenContext(ContextConfig{})
+				defer ctx.Close()
```
gavv, line 227: Please check that err in nil, here and in other tests below.

roc/version_test.go
```diff
@@ -80,3 +81,199 @@ func TestVersion_Parse(t *testing.T) {
...
+			}, // throws err
```
gavv, line 228: What err?

roc/version_test.go
```diff
@@ -80,3 +81,199 @@ func TestVersion_Parse(t *testing.T) {
...
+		},
+		{
+			name: "OpenSender",
+			entrypoint: func() {
+				ctx, _ := OpenContext(ContextConfig{})
+				sender, _ := OpenSender(ctx, SenderConfig{
+					FrameSampleRate: 44100,
+					FrameChannels:   ChannelSetStereo,
+					FrameEncoding:   FrameEncodingPcmFloat,
+				})
```
gavv, line 238: Please use makeSenderConfig().

roc/version_test.go
```diff
@@ -80,3 +81,199 @@ func TestVersion_Parse(t *testing.T) {
...
+		},
+		{
+			name: "OpenReceiver",
+			entrypoint: func() {
+				ctx, _ := OpenContext(ContextConfig{})
+				recv, _ := OpenReceiver(ctx, ReceiverConfig{
+					FrameSampleRate: 44100,
+					FrameChannels:   ChannelSetStereo,
+					FrameEncoding:   FrameEncodingPcmFloat,
+				})
```
gavv, line 251: Please use makeReceiverConfig().

roc/version_test.go
```diff
@@ -80,3 +81,199 @@ func TestVersion_Parse(t *testing.T) {
...
+				defer recv.Close()
+			},
+		},
+		{
+			name:       "ParseEndpoint",
+			entrypoint: func() { _, _ = ParseEndpoint("rtp://192.168.0.1:1234") },
+		},
+		{
+			name:       "SetLogLevel",
+			entrypoint: func() { SetLogLevel(LogDebug) },
```
gavv, line 262: Please use default log level, or restore default log level.

roc/version_test.go
```diff
@@ -80,3 +81,199 @@ func TestVersion_Parse(t *testing.T) {
...
+					Minor: 1,
+					Patch: 0,
+				},
+			},
+			wantErr: nil,
+		},
+	}
+
+	for _, tt := range tests {
+		err := tt.version.Validate()
```
gavv, line 199: Please use t.Run for each table entry to make it subtest.

roc/version_test.go
```diff
@@ -80,3 +81,199 @@ func TestVersion_Parse(t *testing.T) {
...
+			entrypoint: func() { SetLogger(nil) },
+		},
+		{
+			name:       "SetLoggerFunc",
+			entrypoint: func() { SetLoggerFunc(nil) },
+		},
+	}
+
+	for _, tt := range tests {
+		tt.entrypoint()
```
gavv, line 275: Please use t.Run for each table entry to make it subtest.

gavv (commented):
Thanks for update! Some more comments

roc/version_test.go
fastbyt3: my bad had that comment when i didnt resolve the race condition error

gavv:
Please rebase on fresh `main` to fix CI

gavv:
Thanks for update!

I've pushed a follow-up commit with some additional cosmetic changes: 78625302f3d8c25724e4494c6cfecac03f2e2f05

* reordered some declarations and blocks
* reworded some comments
* reworded error messages a bit
* renamed versionOnce to versionInfoOnce
* fixed TestVersion_Validite: `require.Error(t, err, tt.wantErr)` was not actually checking for error message equality and some subtests actually had incorrect error messages; I replaced the check with `require.Equal(t, err, tt.wantErr)`
* made versionCheckCalled a local variable in subtest
* replaced panics with `require.NoError(t, err)`
* replaced LogError with defaultLogLevel

Issue #70, opened 2023-04-03
Add debug logging


gavv:
Add debug logs for all public methods of Context, Sender, and Receiver, except real-time operations: ReadFloats and WriteFloats.

We should log beginning and end of each operation, so that by looking at logs we can easily say when we entered and exited native code. We can do it by adding something like this to each method (pseudocode):

```
log(...)
defer log(...)
```

We should include arguments passed to methods to logs. For enums, let's use stringer tool to generate String() method. https://pkg.go.dev/golang.org/x/tools/cmd/stringer

During Context opening, we should also log versions of bindings and native lib, reported by Version() call.

Logs should be passed to the handler registered via SetLoggerFunc() with log level set to debug. They should take into account log level set by SetLogLevel().

gavv:
Thanks!

Darrellbor:
@gavv I need a bit more context on this.
I believe the objective is to push the logs through the log handler on `log.go`

A possible log impl might look like 
```go
SetLogLevel(LogDebug)
defer SetLogLevel(defaultLogLevel)

SetLoggerFunc(func(msg LogMessage) {
	//what happens here	
})
defer SetLoggerFunc(nil)
```

and would this be replicated across `context`, `receiver` and `sender`?
I'm probably not seeing it the right way... please advise

gavv:
> I believe the objective is to push the logs through the log handler on log.go

Right.

> A possible log impl might look like

I don't see why you may need to modify level and handler.

What you need is to push log message to the registered handler. Also you need to check what level is set, and don't push message if the level is not high enough (if level is info, you shouldn't push debug messages).

See #74, it reworks log handling a bit. I suggest you to base your work on top of that branch (we'll merge it soon).

In that branch, log handler is invoked in goroutine:

```go
func loggerRoutine() {
	for message := range loggerCh {
		fn := loggerFunc.Load().(LoggerFunc)
		if fn != nil {
			fn(message)
		}
	}
}
```

So what you need is to push your messages to `loggerCh`.

You can add a new function like this:

```go
func logWrite(level, text) {
    // don't forget to check level
    loggerCh <- LogMessage{
        // fill fields from args
    }
}
```

and then use this func from context, sender, receiver, etc.

PR #80, opened 2023-04-19
Add debug logging

Darrellbor:
PR for #70 

Implemented 
- [x] log write functionality that pushes `LogMessage` to the `loggerCh` channel
- [x] Add log write to all non-real-time methods of context
- [x] Add log write to all non-real-time methods of receiver
- [x] Add log write to all non-real-time methods of sender

### DIFF
--- a/roc/log.go
+++ b/roc/log.go
@@ -195,21 +195,7 @@ func logWrite(level LogLevel, text string, params ...interface{}) {
 		return
 	}
 
-	var finalPath string
-	_, file, line, _ := runtime.Caller(1)
-
-	dirs := strings.FieldsFunc(file, func(c rune) bool {
-		return c == '/' || c == '\\'
-	})
-	fileNameIndex := len(dirs) - 1
-	lastDirIndex := len(dirs) - 2
-
-	if lastDirIndex >= 0 {
-		fileName := dirs[fileNameIndex]
-		lastDir := dirs[lastDirIndex]
-
-		finalPath = filepath.Join(lastDir, fileName)
-	}
+	file, line := logLocation(2)
 
 	loggerCh <- LogMessage{
 		Level:  level,
@@ -217,10 +203,33 @@ func logWrite(level LogLevel, text string, params ...interface{}) {
 		Pid:    uint64(os.Getpid()),
 		Tid:    uint64(C.rocGoThreadID()),
 		Module: "roc_go",
-		File:   finalPath,
+		File:   file,
 		Line:   line,
 		Text:   fmt.Sprintf(text, params...),
 	}
+}
+
+func logLocation(stack int) (string, int) {
+	_, file, line, ok := runtime.Caller(stack)
+	if !ok {
+		return "", -1
+	}
+
+	parts := strings.FieldsFunc(file, func(c rune) bool {
+		return c == '/' || c == '\\'
+	})
+	if len(parts) == 0 {
+		return "", -1
+	}
+
+	for i := len(parts) - 1; i != 0; i-- {
+		if parts[i] == "roc" {
+			parts = parts[i:]
+			break
+		}
+	}
+
+	return filepath.Join(parts...), line
 }
 
 func init() {


Darrellbor:
@roc-streaming/go ready for review

gavv:
#74 is merged. Could you please **rebase** on fresh main?

Darrellbor:
> #74 is merged. Could you please **rebase** on fresh main?

done

gavv:
> > #74 is merged. Could you please **rebase** on fresh main?
> 
> done

You actually did not rebase, you did a merge. Commits from another pr are still included in this one.

Darrellbor:
> You actually did not rebase, you did a merge. Commits from another pr are still included in this one.
You are right, I thought you meant merge, I've rebased it

Darrellbor:
@roc-streaming/go ready for review

roc/context.go
```diff
@@ -41,6 +42,11 @@ type Context struct {
 // Allocates and initializes a new context. May start some background threads.
 // User is responsible to call Context.Close to free context resources.
 func OpenContext(config ContextConfig) (*Context, error) {
+	logWrite(LogDebug, fmt.Sprintf("%v", logFormat{op: "roc_context_open_begin()",
+		params: map[string]interface{}{
+			"config":  config,
+			"version": Version(),
+		}}))
```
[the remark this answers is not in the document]
gavv: (3) `roc_context_open_begin` looks like an function name, but there is no such function, and it may be confusing. I suggest just to use current Go function everywhere to keep it simple:

```go
logWrite(LogDebug, "entering OpenContext(): config=%+v", config)
```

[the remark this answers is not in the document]
gavv: (4) `OpenContext` has multiple return paths, so just adding `logWrite` to the end is incomplete. Instead, we need to use defer:

```go
logWrite(LogDebug, "entering OpenContext(): config=%+v", config)
defer logWrite(LogDebug, "leaving OpenContext()")
```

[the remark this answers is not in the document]
gavv: (5) It would be very convenient to log returned error, if it's present. We could use named returns:

```go
func OpenContext(config ContextConfig) (ctx *Context, err error) {
    logWrite(LogDebug, "entering OpenContext(): config=%+v", config)
    defer logWrite(LogDebug, "leaving OpenContext(): err=%v", err)
    ...
```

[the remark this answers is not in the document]
gavv: (6) Finally, it would be convenient to log context / sender / receiver pointer when they're created, closed, and in all methods. This will allow distinguishing multiple instances of same type in logs.

```go
func OpenContext(config ContextConfig) (ctx *Context, err error) {
    logWrite(LogDebug, "entering OpenContext(): config=%+v", config)
    defer logWrite(LogDebug, "leaving OpenContext(): context=%p err=%v", ctx, err)
    ...
```

```go
func OpenSender(context *Context, config SenderConfig) (sender *Sender, err error) {
    logWrite(LogDebug, "entering OpenSender(): context=%p config=%+v", context, config)
    defer logWrite(LogDebug, "leaving OpenSender(): context=%p sender=%p err=%v", ctx, sender, err)
    ...
```

roc/log.go
```diff
@@ -181,6 +192,36 @@ func loggerRoutine() {
...
 }
 
+func getGID() uint64 {
+	b := make([]byte, 64)
+	b = b[:runtime.Stack(b, false)]
+	b = bytes.TrimPrefix(b, []byte("goroutine "))
+	b = b[:bytes.IndexByte(b, ' ')]
+	n, _ := strconv.ParseUint(string(b), 10, 64)
+	return n
+}
```
gavv, line 202: LogMessage TID field is for thread ID, not goroutine ID. I think it's better to keep filling it with TID, not GID, for two reasons:

* to avoid confusion when reading logs, otherwise you don't know who to interpret the integer you see if you don't know if the message originated from C or Go

* to keep track of native threads; we may see in logs what native threads are used in C code, and it may be important to understand if they are the same or different native threads used in Go code (e.g. when a Go callback is invoked from C)

roc/log.go
```diff
@@ -181,6 +192,36 @@ func loggerRoutine() {
...
+
+func logWrite(level LogLevel, text string) {
+	if level >= LogDebug {
```
gavv, line 205: We should check `level` against currently used log level, not just `LogDebug`. For that, we'll need to remember level in `SetLogLevel()`. To avoid data race, we can use atomic.

roc/log.go
```diff
@@ -181,6 +192,36 @@ func loggerRoutine() {
...
+		module, ok := debug.ReadBuildInfo()
```
gavv, line 206: I think we can just hard-code module to "roc_go". It will fit well with other modules used for logging - big internal parts of roc-toolkit, like "roc_core", "roc_audio", "roc_rtp", etc.

roc/log.go
```diff
@@ -181,6 +192,36 @@ func loggerRoutine() {
...
+		_, file, line, callerOk := runtime.Caller(0)
+
+		if ok && callerOk {
```
gavv, line 209: I think we can log message even if callerOk is false. Missing file and line is not that critical.

gavv (commented):
Thanks for PR!

gavv:
Note: don't forget to add version info logging to `versionCheck()`.

roc/log.go
Darrellbor: I get errors flagged for this 
```
./log.go:6:1: error: unknown type name 'uint64_t'
uint64_t rocGoThreadID();
```

```
go list failed to return CompiledGoFiles. This may indicate failure to perform cgo processing; try building at the command line. See https://golang.org/issue/38990.go list
```

gavv: This is just a prototype, I didn't test the code.

Darrellbor: i'm not quite sure how to go about this

roc/log.go
Darrellbor: Can you elaborate a bit more on this

roc/log.go
gavv: No worries, then let's keep TID unset for now, and I'll implement the C part by myself after merging this PR.

roc/log.go
gavv: Your logErite func does tge following check: if given level is higher than debug, then print the message. 

Instead, the check should be: if given level is not higher than the level configured by user via SetLogLevel, then print it.

When the user invokes e.g. SetLogLevel(info), then only error and info messages should be printed, and debug and trace should be dropped.

See comment for SetLogLevel.

roc/log.go
Darrellbor: alright noted

gavv:
Sorry for creating conflicts with version checking code.

roc/context.go
```diff
@@ -40,7 +40,10 @@ type Context struct {
 // Open a new context.
 // Allocates and initializes a new context. May start some background threads.
 // User is responsible to call Context.Close to free context resources.
-func OpenContext(config ContextConfig) (*Context, error) {
+func OpenContext(config ContextConfig) (ctx *Context, err error) {
+	logWrite(LogDebug, "entering roc_context_open(): config=%+v", config)
+	defer logWrite(LogDebug, "leaving roc_context_open(): context=%p err=%v", ctx, err)
```
gavv, line 45: Since these log messages surround theb whole OpenContex, not just the call to native roc_open_context, I suggest to replace roc_open_context with OpenContext in the message.

Same for other methods. Since we're logging the whole go method, let's use go method name.

roc/log.go
Darrellbor: test done

roc/context.go
Darrellbor: done

roc/receiver.go
```diff
@@ -171,7 +171,12 @@ type Receiver struct {
 
 // Open a new receiver.
 // Allocates and initializes a new receiver, and attaches it to the context.
-func OpenReceiver(context *Context, config ReceiverConfig) (*Receiver, error) {
+func OpenReceiver(context *Context, config ReceiverConfig) (receiver *Receiver, err error) {
+	logWrite(LogDebug, "entering OpenReceiver(): config=%+v context=%p", config, context)
```
gavv, line 175: nit: everywhere we log context, let's make it the first in the log message, for consistency.

roc/receiver.go
```diff
@@ -237,7 +242,12 @@ func OpenReceiver(context *Context, config ReceiverConfig) (*Receiver, error) {
 // calling Receiver.Connect() for the interface.
 //
 // Automatically initializes slot with given index if it's used first time.
-func (r *Receiver) SetMulticastGroup(slot Slot, iface Interface, ip string) error {
+func (r *Receiver) SetMulticastGroup(slot Slot, iface Interface, ip string) (err error) {
+	logWrite(LogDebug,
+		"entering Receiver.SetMulticastGroup(): slot=%v iface=%v ip=%v", slot, iface, ip,
+	)
```
gavv, line 248: nit: missing receiver/sender %p in "entering" message here and in many other places.

roc/receiver.go
```diff
@@ -171,7 +171,12 @@ type Receiver struct {
 
 // Open a new receiver.
 // Allocates and initializes a new receiver, and attaches it to the context.
-func OpenReceiver(context *Context, config ReceiverConfig) (*Receiver, error) {
+func OpenReceiver(context *Context, config ReceiverConfig) (receiver *Receiver, err error) {
```
gavv, line 174: Here and in most other places, you introduced named returns, but did not modify the code to use them. In result, defer logs variables that were never set. Please test your code when you're submitting PR for review.

roc/version.go
```diff
@@ -84,8 +84,12 @@ func (vi VersionInfo) Validate() error {
 }
 
 func checkVersion() {
+	var vi VersionInfo
+	logWrite(LogDebug, "entering versionCheck()")
+	defer logWrite(LogDebug, "leaving versionCheck(): version=%+v", vi)
```
gavv, line 89: There is no need to log every checkVersion call because in all invocations except first it's no-op.

Instead, it would be enough to log the retrieved version only once, after fetchVersion call and before validating it.

roc/log.go
```diff
@@ -147,6 +149,7 @@ func (standardLogger) Print(v ...interface{}) {
 }
 
 var (
+	logLevel   atomic.Value
```
gavv, line 152: For consistency with other vars, let's rename it to `loggerLevel`.

There is no need to use atomic.Value here because it's just an int. You can use an int of an appropriate size instead, and corresponding atomic functions.

roc/log.go
```diff
@@ -182,6 +185,24 @@ func loggerRoutine() {
...
+	if currentLogLevel != nil {
+		if currentLogLevel := currentLogLevel.(LogLevel); level <= currentLogLevel {
+			loggerCh <- LogMessage{
+				Level: level,
+				Time:  time.Now(),
+				Pid:   uint64(os.Getpid()),
+				//Tid:    -get from C module-
+				Module: "roc_go",
+				File:   "log.go",
+				Line:   210,
```
gavv, line 199: I guess you hard-coded file and line accidentally? Obtaining them from stack, as it was in your previous version of the patch, was a good implementation.

roc/log.go
```diff
@@ -182,6 +185,24 @@ func loggerRoutine() {
...
+				Text:   fmt.Sprintf("%+v", text),
```
gavv, line 200: Did you test it? You're not properly formatting message text.

roc/log_test.go
```diff
@@ -140,6 +140,40 @@ func TestLog_Func(t *testing.T) {
 	}
 }
 
+func TestLog__LogWrite(t *testing.T) {
```
gavv, line 143: Extra `_`

roc/log_test.go
```diff
@@ -140,6 +140,40 @@ func TestLog_Func(t *testing.T) {
...
+
+	select {
+	case msg := <-ch:
+		assert.Equal(t, LogDebug, msg.Level, "Expected log level to be debug")
+		assert.NotEmpty(t, msg.Module)
+		assert.Equal(t, "roc_go", msg.Module)
+		assert.NotEmpty(t, msg.File)
+		assert.Equal(t, "log.go", msg.File)
+		assert.NotEmpty(t, msg.Line)
+		assert.Equal(t, 210, msg.Line)
```
gavv, line 171: nit: Could you also please check time, pid, tid, text, as I described earlier?

gavv:
I've created PR for retrieving TID: https://github.com/roc-streaming/roc-go/pull/93

roc/log.c
```diff
@@ -3,4 +3,4 @@
 
 void rocGoLogHandlerProxy(const roc_log_message* message, void* argument) {
     rocGoLogHandler((roc_log_message*)message);
-}
+}
```
gavv, line 6: Please rollback this change.

gavv:
> I've created PR for retrieving TID: #93

Merged. Could you please rebase and use new function? (`C.rocGoThreadID`) to fill TID?

roc/log_test.go
Darrellbor: I added tests for all these except the exact value for Tid since it's gotten from the C module and I can't use that within the testing area, so instead I tested for nullability

roc/log.c
Darrellbor: what change?

gavv: Huh, github does not show it in all views. If you open https://github.com/roc-streaming/roc-go/pull/80/files, you can see it:

image

It means that `\n` was removed from the end of the file, in other words, the last line in the file now does not have trailing newline, unlike all other lines. Probably your editor did that.

roc/receiver.go
```diff
@@ -171,18 +171,25 @@ type Receiver struct {
...
+	logWrite(LogDebug, "entering OpenReceiver(): context=%+v config=%p", context, config)
+	defer logWrite(LogDebug,
+		"leaving OpenReceiver(): context=%p receiver=%p err=%v", context, receiver, err,
+	)
```
gavv, line 178: In OpenSender and OpenReceiver, `%+v` and `%p` are messed up.

roc/receiver.go
```diff
@@ -316,16 +341,23 @@ func (r *Receiver) SetReuseaddr(slot Slot, iface Interface, enabled bool) error
 // If endpoint has explicitly set zero port, the receiver is bound to a randomly
 // chosen ephemeral port. If the function succeeds, the actual port to which the
 // receiver was bound is written back to endpoint.
-func (r *Receiver) Bind(slot Slot, iface Interface, endpoint *Endpoint) error {
+func (r *Receiver) Bind(slot Slot, iface Interface, endpoint *Endpoint) (err error) {
+	logWrite(LogDebug,
+		"entering Receiver.Bind(): receiver=%p slot=%+v iface=%+v endpoint=%+v", r, slot, iface, endpoint,
+	)
+	defer logWrite(LogDebug, "leaving Receiver.Bind(): receiver=%p err=%v", r, err)
```
gavv, line 348: Since Bind() modifies endpoint argument, let's log endpoint in defer statement as well.

roc/version.go
```diff
@@ -87,6 +87,9 @@ func checkVersion() {
 	if atomic.CompareAndSwapInt32(&versionCheckOnce, 0, 1) {
 		vi := fetchVersion()
 
+		logWrite(LogDebug, "entering versionCheck()")
+		defer logWrite(LogDebug, "leaving versionCheck(): version=%+v", vi)
```
gavv, line 91: As I already wrote before, there is no need for entering/leaving message pair here, we can just log version after fetching it and before validating it. There is no big work done between enter and leave, unlike other methods.

Also using defer hurts here, because this way Validate() will panic before version is logged.

roc/log.go
```diff
@@ -183,6 +187,23 @@ func loggerRoutine() {
 	}
 }
 
+func logWrite(level LogLevel, text ...interface{}) {
+	if level <= loggerLevel {
```
gavv, line 191: This is a data race because you're reading loggerLevel in logWrite and writing it in SetLogLevel concurrently without protections.

As I already wrote, you can make loggerLevel a plain integer, but you still need to use appropriatre atomic operations to work with it.

roc/log.go
```diff
@@ -183,6 +187,23 @@ func loggerRoutine() {
...
+		_, file, line, _ := runtime.Caller(0)
```
gavv, line 192: It seems we're running circles here.

As I already wrote, this will return file and line that corresponds to `log.go`. In other words, file and line will always be `log:go` and `192`.

What we need instead is to obtain the file and line which correspond to the place from which `logWrite()` was invoked.

roc/log_test.go
```diff
@@ -140,6 +141,47 @@ func TestLog_Func(t *testing.T) {
...
+
+	ctx, err := OpenContext(ContextConfig{})
+	require.NoError(t, err)
+	ctx.Close()
+
+	select {
+	case msg := <-ch:
+		assert.Equal(t, LogDebug, msg.Level, "Expected log level to be debug")
+		assert.NotEmpty(t, msg.Module)
+		assert.Equal(t, "roc_go", msg.Module)
```
gavv, line 168: NotEmpty for Module is redundant because we already have Equal check.

roc/log_test.go
```diff
@@ -140,6 +141,47 @@ func TestLog_Func(t *testing.T) {
...
+		assert.NotEmpty(t, msg.File)
+		assert.Equal(t, "log.go", msg.File[len(msg.File)-6:])
```
gavv, line 170: After you fix file & line filling, `File` should become "context.go" in this test. NotEmpty is redundant.

roc/log_test.go
```diff
@@ -140,6 +141,47 @@ func TestLog_Func(t *testing.T) {
...
+		assert.NotEmpty(t, msg.Line)
+		assert.Equal(t, 192, msg.Line)
```
gavv, line 172: Hard-coding specific line is not very good, it may frequently change due to modifications not related to this test, and it should not break the test.

Instead of NotEmpty + Equal, let's just check that line is positive number.

roc/log_test.go
```diff
@@ -140,6 +141,47 @@ func TestLog_Func(t *testing.T) {
...
+		assert.NotEmpty(t, msg.Pid)
+		assert.Equal(t, uint64(os.Getpid()), msg.Pid)
```
gavv, line 174: NotEmpty check is redundant here

roc/log_test.go
```diff
@@ -140,6 +141,47 @@ func TestLog_Func(t *testing.T) {
...
+		assert.NotEmpty(t, msg.Tid)
+		assert.NotEmpty(t, msg.Time)
+		assert.Equal(t, time.Now().Truncate(time.Minute), msg.Time.Truncate(time.Minute))
```
gavv, line 177: * NotEmpty is redundant
* A more natural and correct check would be to remember time at the start of the test and check that the message time is not before that remembered time and not after now

roc/log_test.go
```diff
@@ -140,6 +141,47 @@ func TestLog_Func(t *testing.T) {
...
+		assert.NotEmpty(t, msg.Text)
+		assert.Equal(t, "entering OpenContext(): config=%+v%!(EXTRA roc.ContextConfig={0 0})", msg.Text)
```
gavv, line 179: * NotEmpty is redundant
* I think equality check is too strict here, we can just do some sanity check, e.g. ensure that text contains "OpenContext" or maybe "entering OpenContext"

gavv:
Sorry for asking this again, but did you self-review and test your code before submitting it? I mean, enable logging and actually verify the logger output?

Formatting is still broken; the messages produced by logWrite are a mess.

This is a bit of wasting time of both of us, because I need to debug it locally, write a comment, then you need to read my comment, and debug it locally. So we're doubling debugging work and spend extra time for communication.

gavv:
Regarding named returns. The current code looks a bit messy: it mixes named returns with regular returns like this:

```go
	if endpoint == nil {
		err = errors.New("endpoint is nil")
		return err
	}
```

In other words, we first fill named return var `err`, but then anyway use regular `return err`.

To keep code easier to follow, I suggest to completely switch to **naked** returns everywhere where we use logWrite, like this:

```go
	if endpoint == nil {
		err = errors.New("endpoint is nil")
		return
	}
```

and:

```go
err = nil
return
```

gavv:
We forgot to update 1 place to use named return vars:
https://github.com/roc-streaming/roc-go/blob/61892242983ae2ca830c6ff743160071569ab8a6/roc/context.go#L56C19-L58

gavv:
For consistency, I suggest to update ReadFloats and WriteFloats to use named return vars + naked returns too.

roc/log.c
Darrellbor: > Sorry for asking this again, but did you self-review and test your code before submitting it? I mean, enable logging and actually verify the logger output?
> 
> Formatting is still broken; the messages produced by logWrite are a mess.
> 
> This is a bit of wasting time of both of us, because I need to debug it locally, write a comment, then you need to read my comment, and debug it locally. So we're doubling debugging work and spend extra time for communication.

I'm sorry for this, but can i ask how i can enable logging and actually test it on my end... If there's a specific way that you do this

Darrellbor:
> Sorry for asking this again, but did you self-review and test your code before submitting it? I mean, enable logging and actually verify the logger output?
> 
> Formatting is still broken; the messages produced by logWrite are a mess.
> 
> This is a bit of wasting time of both of us, because I need to debug it locally, write a comment, then you need to read my comment, and debug it locally. So we're doubling debugging work and spend extra time for communication.

I'm sorry for this, but can i ask how i can enable logging and actually test it on my end... If there's a specific way that you do this

gavv:
You can just hack log.go and maybe main_test.go to ensure logs are not disabled in tests.

gavv:
> To keep code easier to follow, I suggest to completely switch to **naked** returns everywhere where we use logWrite, like this:
> 
> ```go
> 	if endpoint == nil {
> 		err = errors.New("endpoint is nil")
> 		return
> 	}
> ```

I have created a small linter that will help us to maintain this code style: https://github.com/gavv/returnstyles

You can install it to GOPATH using:

```
go install github.com/gavv/returnstyles@latest
```

And then run like this:

```
cd roc
returnstyles -named-allow=true -named-allow-normal-returns=false .
```

(see linter README for explanations)

It will report messages like:

```
/home/victor/dev/roc-streaming/roc-go/roc/context.go:57:3: normal (non-naked) returns not allowed in functions with named return variables
/home/victor/dev/roc-streaming/roc-go/roc/context.go:67:2: normal (non-naked) returns not allowed in functions with named return variables
/home/victor/dev/roc-streaming/roc-go/roc/context.go:85:4: normal (non-naked) returns not allowed in functions with named return variables
/home/victor/dev/roc-streaming/roc-go/roc/context.go:91:2: normal (non-naked) returns not allowed in functions with named return variables
```

It should help you to find all problematic places that should be fixed. Later we'll integrate this linter into makefile and CI.

Darrellbor:
@gavv ready for re-review

gavv:
```
2023/05/22 22:28:34 [dbg] roc_go: entering OpenContext(): config={MaxPacketSize:0 MaxFrameSize:0}
...
2023/05/22 22:28:34 [dbg] roc_go: leaving OpenContext(): context=0x0 err=<nil>
```

We have a problem - context is 0x0 in "leaving" message. See explanation here: https://stackoverflow.com/questions/42703707/

To fix it, we can convert this:

```go
	logWrite(LogDebug, "entering OpenContext(): config=%+v", config)
	defer logWrite(LogDebug, "leaving OpenContext(): context=%p err=%v", ctx, err)
```

to this:

```go
	logWrite(LogDebug, "entering OpenContext(): config=%+v", config)
	defer func() {
		logWrite(LogDebug, "leaving OpenContext(): context=%p err=%v", ctx, err)
	}()
```

Could you please apply this fix everywhere and test that everything is logged correctly?

gavv:
Everywhere where we log errors, I suggest to use `err=%#v` instead of `err=%v`.

With `%v`:

```
2023/05/22 23:06:25 [dbg] roc_go: leaving Receiver.SetMulticastGroup():
  receiver=0xc00001e6a0 err=roc_receiver_set_multicast_group() failed with code -1
```

With `%#v`:

```
2023/05/22 23:06:47 [dbg] roc_go: leaving Receiver.SetMulticastGroup():
  receiver=0xc000172500 err=roc.nativeErr{op:"roc_receiver_set_multicast_group()", code:-1}
```

These logs are intended to be read by bindings developers, and I think for this purpose `%#v` is more informative.

gavv:
There is a difference how File field is formatted by bindings and native lib:

```
2023/05/22 23:30:24 [dbg] roc_go: /home/victor/dev/roc-streaming/roc-go/roc/context.go:44
  entering OpenContext(): config={MaxPacketSize:0 MaxFrameSize:0}
```

```
2023/05/22 23:30:24 [dbg] roc_core: src/internal_modules/roc_core/slab_pool.cpp:38
  slab pool: initializing: object_size=672 min_slab=0B(1S) max_slab=0B(0S) poison=0
```

For bindings, it's an absolute path: `/home/victor/dev/roc-streaming/roc-go/roc/version.go`

For native lib, it's a path relative to project root: `src/internal_modules/roc_core/slab_pool.cpp`

It would be nice to unify formatting and make bindings path relative to project root too, i.e. instead of `/home/victor/dev/roc-streaming/roc-go/roc/version.go` show `roc/version.go`.

We can achieve this by finding the most right `/roc/` component in the path and truncating everything that is on the left of it.

Note: we should do it in a cross-platform way, i.e. we should not rely on specific path separator (`/` or `\`).

gavv:
(I think we should do it in logWrite)

gavv:
If this commit: https://github.com/roc-streaming/roc-go/pull/80/commits/4fef44facb574d98aae4f85061534a02873fa8fa

Is ready for review and is intended to fix this: https://github.com/roc-streaming/roc-go/pull/80#issuecomment-1557835660

Then it's not the algorithm I suggested. This implementation makes an assumption about folder nesting and will break if it's changed somehow (e.g. we add subdirs).

I suggested a more relaxed requirement which will work with any nesting.

Other ideas are welcome too.

gavv:
PR for adding `returnstyles` linter to repo: https://github.com/roc-streaming/roc-go/pull/95

Darrellbor:
> If this commit: [4fef44f](https://github.com/roc-streaming/roc-go/commit/4fef44facb574d98aae4f85061534a02873fa8fa)
> 
> Is ready for review and is intended to fix this: [#80 (comment)](https://github.com/roc-streaming/roc-go/pull/80#issuecomment-1557835660)
> 
> Then it's not the algorithm I suggested. This implementation makes an assumption about folder nesting and will break if it's changed somehow (e.g. we add subdirs).
> 
> I suggested a more relaxed requirement which will work with any nesting.
> 
> Other ideas are welcome too.

This was one way of doing it, alternatively i thought of splitting by `\` and `/` and joining the last two together but felt the path route would be platform agnostic.

roc/log_test.go
```diff
@@ -140,6 +141,49 @@ func TestLog_Func(t *testing.T) {
...
+			assert.NotEmpty(t, msg.Module)
+			assert.Equal(t, "roc_go", msg.Module)
+			assert.Equal(t, "context.go", msg.File[len(msg.File)-10:])
+			assert.Greater(t, msg.Line, 0, "Line number must be positive")
+			assert.Equal(t, uint64(os.Getpid()), msg.Pid)
+			assert.NotEmpty(t, msg.Tid)
+			assert.True(t,
+				msg.Time.After(testStartTime),
+				"Time assertion failed: test time is less than start time of the test",
+			)
```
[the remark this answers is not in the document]
Darrellbor: I introduced a tolerance of a millisecond to compare the time to also account for when the time is the same

roc/log_test.go
```diff
@@ -140,6 +141,49 @@ func TestLog_Func(t *testing.T) {
...
+		}
+	})
+	defer SetLoggerFunc(nil)
+
+	ctx, err := OpenContext(ContextConfig{})
+	require.NoError(t, err)
+
+	select {
+	case msg := <-ch:
+		if msg.Module == "roc_go" && strings.Contains(msg.Text, "entering OpenContext()") {
```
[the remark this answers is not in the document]
Darrellbor: I tried removing this but I was getting `leaving Context.Close()` being pushed into the channel instead

gavv:
> This was one way of doing it, alternatively i thought of splitting by \ and / and joining the last two together but felt the path route would be platform agnostic.

Joining **the last two** will lead to the same problem: we're hard-coding fixed nesting level (2); I'd like to avoid it, it's very implicit and not very flexible.

gavv:
You could use

roc/log_test.go
```diff
@@ -180,3 +181,51 @@ func TestLog_Levels(t *testing.T) {
...
+		assert.True(t,
+			msg.Time.After(testStartTime.Add(-time.Millisecond)) &&
+				msg.Time.Before(testStartTime.Add(time.Millisecond)),
+			"Time assertion failed: test time is not within the tolerance of the start time of the test",
+		)
+		assert.True(t,
+			msg.Time.After(time.Now().Add(-time.Millisecond)) &&
+				msg.Time.Before(time.Now().Add(time.Millisecond)),
+			"Time assertion failed: message time cannot be greater than the time now",
+		)
```
gavv, line 224:
```suggestion
		assert.WithinRange(t, msg.Time, testStartTime, time.Now(),
			"Time must have meaningful value")
```

gavv: This way there is no need for magic millisecond because the range is inclusive. Also the code is much shorter.

Darrellbor: WithinRange doesn't seem to exist on the assert package

gavv:
I've created issue to allow enabling logs in tests: https://github.com/roc-streaming/roc-go/issues/96

Darrellbor:
> You could use https://pkg.go.dev/path/filepath#SplitList

I attempted using this but it just returns the path in full
I use it after using the `filepath.ToSlash(file)` method

gavv:
A guy on reddit [suggested](https://www.reddit.com/r/golang/comments/13njcs0/comment/jl9jl5a/?utm_source=share&utm_medium=web2x&context=3) that naked returns are not actually needed here.

Earlier I did an incorrect test and decided that we need naked returns. But in fact, if we properly wrap deferred code in `func(){}`, as we already did recently, we can still use normal (non-naked) returns and return variables will have desired values in defer:

```go
	logWrite(LogDebug, "entering OpenContext(): config=%+v", config)
	defer func() {
		logWrite(LogDebug, "leaving OpenContext(): context=%p err=%#v", ctx, err)
	}()

	...

	if errCode != 0 {
		return nil, newNativeErr("roc_context_open()", errCode)
	}
```

The code is cleaner with non-naked returns, so could we please switch back to using them?

Sorry for the mess!

gavv:
> > You could use https://pkg.go.dev/path/filepath#SplitList
> 
> I attempted using this but it just returns the path in full I use it after using the `filepath.ToSlash(file)` method

Oh, you're right, SplitList is for splitting colon-separated list of paths (like `PATH`), not for splitting single path into components.

Well then I guess you could use filepath.Split iteratively until you reach "roc" parent directory, or something like this.

roc/log_test.go
Darrellbor: assert.WithinRange doesn't exist though

Darrellbor:
```
if errCode != 0 {
		return nil, newNativeErr("roc_context_open()", errCode)
	}
```
Wouldn't i need to still assign the values of `ctx` and `err`?

gavv:
> ```
> if errCode != 0 {
> 		return nil, newNativeErr("roc_context_open()", errCode)
> 	}
> ```
> 
> Wouldn't i need to still assign the values of `ctx` and `err`?

I also suggest to expand logWrite test a bit, and check not only "enetering" but also "leaving" message and check that the leaving message contains "context=0x..." where "..." is not zero address.

Darrellbor:
@gavv 
I have worked on the following
- extended the test to accommodate `leaving OpenContext()` as well as checking that the `context=0x...` is accounted for
- reverted back to non-naked returns style for all the previously changed methods
- reverted back to the method I used in the first place as `filepath.Split` didn't do what I intended and would result in the same approach for implementation.
- assert.WithinRange doesn't exist so I reverted back to the assert.True methods

roc/log_test.go
```diff
@@ -180,3 +182,56 @@ func TestLog_Levels(t *testing.T) {
...
+			assert.Contains(t, msg.Text, "OpenContext()")
+			if i == 1 {
+				assert.Contains(t, msg.Text, fmt.Sprintf("OpenContext(): context=%p", ctx))
+			}
+		case <-time.After(time.Minute):
+			t.Fatal("expected logs, didn't get them before timeout")
+		}
+	}
+
+	ctx.Close()
```
gavv, line 236: nit: better use `defer ctx.Close()` after creating context.

roc/receiver.go
```diff
@@ -237,17 +244,24 @@ func OpenReceiver(context *Context, config ReceiverConfig) (*Receiver, error) {
...
 		return errors.New("receiver is closed")
 	}
 
-	cIP, err := go2cStr(ip)
-	if err != nil {
-		return fmt.Errorf("invalid ip: %w", err)
+	cIP, parseErr := go2cStr(ip)
+	if parseErr != nil {
+		return fmt.Errorf("invalid ip: %w", parseErr)
 	}
```
gavv, line 265: nit: parseErr can be renamed back to err.

roc/receiver.go
```diff
@@ -377,7 +409,7 @@ func (r *Receiver) Bind(slot Slot, iface Interface, endpoint *Endpoint) error {
 //
 // Until the receiver is connected to at least one sender, it produces silence.
 // If the receiver is connected to multiple senders, it mixes their streams into one.
-func (r *Receiver) ReadFloats(frame []float32) error {
+func (r *Receiver) ReadFloats(frame []float32) (err error) {
```
gavv, line 412: nit: named return can be reverted.

roc/log.go
```diff
@@ -183,6 +189,40 @@ func loggerRoutine() {
...
+	})
+	fileNameIndex := len(dirs) - 1
+	lastDirIndex := len(dirs) - 2
+
+	if lastDirIndex >= 0 {
+		fileName := dirs[fileNameIndex]
+		lastDir := dirs[lastDirIndex]
+
+		finalPath = filepath.Join(lastDir, fileName)
+	}
```
gavv, line 212: This is the same algorithm that you already implemented in previous versions, just expressed in different way. You're still implicitly hard-coding specific nesting level (2). Please read carefully what algorithm I suggested as an alternative and why, or feel free to suggest other algorithms that avoid hard-coding nesting level. https://github.com/roc-streaming/roc-go/pull/80#issuecomment-1557835660

gavv (commented):
Thanks for update

roc/log.go
```diff
@@ -183,6 +189,40 @@ func loggerRoutine() {
...
+		Level:  level,
+		Time:   time.Now(),
+		Pid:    uint64(os.Getpid()),
+		Tid:    uint64(C.rocGoThreadID()),
+		Module: "roc_go",
+		File:   finalPath,
+		Line:   line,
+		Text:   fmt.Sprintf(text, params...),
+	}
+}
```
gavv, line 224:
```suggestion
func logWrite(level LogLevel, text string, params ...interface{}) {
	storedLogLevel := LogLevel(atomic.LoadInt32(&loggerLevel))
	if level > storedLogLevel {
		return
	}

	file, line := logLocation(2)

	loggerCh <- LogMessage{
		Level:  level,
		Time:   time.Now(),
		Pid:    uint64(os.Getpid()),
		Tid:    uint64(C.rocGoThreadID()),
		Module: "roc_go",
		File:   file,
		Line:   line,
		Text:   fmt.Sprintf(text, params...),
	}
}

func logLocation(stack int) (string, int) {
	_, file, line, ok := runtime.Caller(stack)
	if !ok {
		return "", -1
	}

	parts := strings.FieldsFunc(file, func(c rune) bool {
		return c == '/' || c == '\\'
	})
	if len(parts) == 0 {
		return "", -1
	}

	for i := len(parts) - 1; i != 0; i-- {
		if parts[i] == "roc" {
			parts = parts[i:]
			break
		}
	}

	return filepath.Join(parts...), line
}
```

gavv: Example implementation of the algorithm I was talking about.

Issue #85, opened 2023-05-13
Use time.Time for LogMessage.Time


gavv:
LogMessage.Time encodes absolute time, in form of nanoseconds since epoch, as uint64. It would be more convenient to encode it as time.Time instead. Comments and tests should be adjusted accordingly.

ParasRaba155:
Hey @roc-streaming/roc-go this is my first time working on an open source project, and I am very novice at that, I have raised a PR and hope to hear your reviews for the PR

PR #89, opened 2023-05-16
Issue 85: Use time.Time for encoding LogMessage.Time instead of uint64

ParasRaba155:
- PR for #85 

- CHANGED the LogMeesage.Time encoding from uint64 to time.Time and And adjusted the testcases to check if time is not zero

- Removed the redudant error check in reciver.go as linter suggested

### DIFF
--- a/roc/log.go
+++ b/roc/log.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"log"
 	"sync/atomic"
+	"time"
 )
 
 // LogLevel defines the logging verbosity.
@@ -49,8 +50,8 @@ type LogMessage struct {
 	// Line number in the source code file.
 	Line int
 
-	// Message timestamp, nanoseconds since Unix epoch.
-	Time uint64
+	// Message timestamp, unix time at which message was logged
+	Time time.Time
 
 	// Platform-specific process ID.
 	Pid uint64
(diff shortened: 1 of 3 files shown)


ParasRaba155:
> AFAIK, we don't have such linter enabled, 
 
as for the linter, that was the `golangci-lint` suggestion on `make lint` command, but that might be due to different version of `golangci-lint` but I will roll back that and also, can you specify me the `golangcli-lint` version that is used in the project so that I can verify mine.

gavv:
Oh, I see, thanks.

CI uses 1.51.2
 https://github.com/roc-streaming/roc-go/blob/main/.github/workflows/build.yaml#L109

What version do you use?

If you're using a later one, I guess we'll need to adjust the linter config and bump it on CI, too. If you're using an older one, you can just update yours.

ParasRaba155:
> CI uses 1.51.2 https://github.com/roc-streaming/roc-go/blob/main/.github/workflows/build.yaml#L109

Mine is 1.52.2 but it is built with go1.20.2, as I have that version of go installed locally and I forgot to install linter from the go1.13

gavv:
A fix for linter was merged in #90.

BTW, welcome to matrix chat of the project! https://roc-streaming.org/toolkit/docs/about_project/contacts.html#matrix-chats

Issue #86, opened 2023-05-13
Add CI step that checks that gofmt was called


gavv:
We can use `Jerome1337/gofmt-action` for that with `gofmt-flags` set to `-s -l`.

gavv:
Landed, thanks @wataru-nakanishi

PR #91, opened 2023-05-20
add gofmt step to ci

wataru-nakanishi:
PR for https://github.com/roc-streaming/roc-go/issues/86

Implemented
- [x]  Add gofmt CI step

This is my first contributing!
If the contributing process was wrong, please tell me.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -71,6 +71,12 @@ jobs:
         with:
           go-version: ${{ matrix.go }}
 
+      - name: gofmt
+        uses: Jerome1337/gofmt-action@v1.0.5
+        with:
+          gofmt-path: './roc'
+          gofmt-flags: '-s -l -d'
+
       - name: Build Roc
         run: |
           git clone https://github.com/roc-streaming/roc-toolkit.git /tmp/roc
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -165,12 +165,26 @@ jobs:
         run: |
           cd roc
           go test -race
+  
+  formatting:
+    runs-on: ubuntu-latest
+
+    name: Check formatting
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: gofmt
+        uses: Jerome1337/gofmt-action@v1.0.5
+        with:
+          gofmt-path: './roc'
+          gofmt-flags: '-s -l -d'
 
   release:
     runs-on: ubuntu-latest
 
     if: ${{ startsWith(github.ref, 'refs/tags/v') }}
-    needs: [ubuntu, macos]
+    needs: [ubuntu, macos, formatting]
 
     name: Check version
     steps:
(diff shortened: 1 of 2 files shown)


wataru-nakanishi:
@roc-streaming/go ready for review

.github/workflows/build.yaml
```diff
@@ -71,6 +71,12 @@ jobs:
         with:
           go-version: ${{ matrix.go }}
 
+      - name: gofmt
```
wataru-nakanishi, line 74: I added fmt step only in linux.
Because Jerome1337/gofmt-action@v1.0.5 is available in linux only.

gavv:
Thanks for PR! Using linux is perfectly ok, we don't need to check formatting on each OS because results would be the same. 

I suggest extracting formatting check into a separate job "Formatting". It's not really bound to any steps in "Linux" job and can be done separately.

gavv:
Follow-up issue: https://github.com/roc-streaming/roc-go/issues/92

gavv:
@wataru-nakanishi Could you please check if this CI step will fail if a file has no newline at the end? If it won't, I think I'll create a separate issue for adding such step too.

wataru-nakanishi:
@gavv 
Thank you!
Fixed.

> Could you please check if this CI step will fail if a file has no newline at the end? If it won't, I think I'll create a separate issue for adding such step too.

I added a new line to config.go and CI failed.
Originally, there are no new line
https://github.com/roc-streaming/roc-go/actions/runs/5035614704/jobs/9031245933?pr=91

gavv:
> I added a new line to config.go and CI failed.
Originally, there is no new line


I actually mean the opposite case: not when extra newline is added, but when the last one is missing.

wataru-nakanishi:
@gavv 
CI check didn't fail if a file has no newline at the end.
For example, config.go has no new line.
https://github.com/roc-streaming/roc-go/blob/main/roc/config.go#L494

Please let me know if I'm misunderstanding.

gavv:
config.go _does_ have newline at the end. It's last symbol is LF, not "}". If there were no terimnating LF, github would display that red icon that I showed above.

wataru-nakanishi:
@gavv 
> config.go does have newline at the end. It's last symbol is LF, not "}". If there were no terimnating LF, github would display that red icon that I showed above.

Thanks for letting me know.
CI failed if a file has no new line.
https://github.com/roc-streaming/roc-go/actions/runs/5035923474/jobs/9031753001?pr=91

gavv:
Thanks!! So we don't need one more check, gofmt already checks everything we need.

One more comment: could you please place formatting job before release job, and add formatting job to the list of dependencies of release job?

wataru-nakanishi:
@gavv 
> could you please place formatting job before release job, and add formatting job to the list of dependencies of release job?

fixed! [38053ff](https://github.com/roc-streaming/roc-go/pull/91/commits/38053ff2950aab249decad1ebe2f72e74efd2ac5)

gavv:
LGTM

Issue #83, opened 2023-05-13
Use time.Duration for time intervals


gavv:
Find fields in SenderConfig and ReceiverConfig that hold time intervals, and switch their type from plane integer to time.Duration. Adjust comments and implementation accordingly.

Fields of interest:

* SenderConfig:

  * [x] PacketLength

* ReceiverConfig

  * [x] TargetLatency
  * [x] MaxLatencyOverrun
  * [x] MaxLatencyUnderrun
  * [x] NoPlaybackTimeout
  * [x] BrokenPlaybackTimeout
  * [x] BreakageDetectionWindow

To avoid copy-paste in implementation, we'll need to add new conversion functions (between time.Duration and C.longlong / C.ulonglong) to convert.go.

We'll need separate functions for conversion to signed and unsigned longs. (unsigned version should return error when it encounters negative duration). New functions should be covered with tests in convert_test.go.

g41797:
i am on board

gavv:
@g41797 makes sense. Just send a PR  so we can see the actual code when discussing this.

g41797:
After adding of
    
     import "C"

to convert_test.go:

[g41797@BLKF roc]$ go build .
use of cgo in test convert_test.go not supported

According to [import "C" is unsupported in test](https://stackoverflow.com/questions/27930737/import-c-is-unsupported-in-test-looking-for-alternatives) possible solution is "...to put the cgo code in a non-test file..."

wrapper wrapgo2cUlonglongs was added to convert.go

g41797:
Commits of PR - removed 
Fork synced with main 
I am going to merge my development with new sources

PR #94, opened 2023-05-21
Commit of "Use time.Duration for time intervals"

g41797:
Done according to former issue #83


roc/config.go
```diff
@@ -379,8 +380,8 @@ type SenderConfig struct {
 	// The samples written to the sender are buffered until the full packet is
 	// accumulated or the sender is flushed or closed. Larger number reduces
 	// packet overhead but also increases latency.
-	// If zero, default value is used.
-	PacketLength uint64
+	// If zero, default value is used. If negative, the error is returned.
```
gavv, line 383: I suggest to rephrase this:

> If negative, the error is returned.

to:

> Should not be negative, otherwise an error is returned.

To make it clear that negative values are not desired usage.

roc/convert.go
```diff
@@ -37,3 +39,34 @@ func go2cBool(b bool) C.uint {
 	}
 	return 0
 }
+
+func go2cUlonglongs(tds map[string]time.Duration) (map[string]C.ulonglong, error) {
```
gavv, line 43: I think using map looks over-complicated and someone who reads the code will spend time wondering why it's here.

Let's instead create functions for single value:

```go
func go2cUnsignedDuration(d time.Duration) ulonglong
func c2goUnsignedDuration(d ulonglong) time.Duration
```

No need for any wrapper function, just create an alias like we do it for `C.char` at the top of this file:

```go
type ulonglong = C.ulonglong
```

No need to pass field name to those functions; instead, please add field name to the error on the caller side, like we do for go2cStr (find its usages in code)

gavv (commented):
Thanks for PR! Here are a couple of comments

roc/convert.go
g41797: Morgen! Morgen! Nur nicht heute!

roc/convert.go
```diff
@@ -37,3 +39,13 @@ func go2cBool(b bool) C.uint {
 	}
 	return 0
 }
+
+type ulonglong = C.ulonglong
```
gavv, line 43: nit: could you please move this to the top, like this?

```go
type (
    char = C.char
    ulonglong = C.ulonglong
)
```

roc/convert.go
```diff
@@ -37,3 +39,13 @@ func go2cBool(b bool) C.uint {
...
+
+func go2cUnsignedDuration(d time.Duration) (ulonglong, error) {
+	if d < 0 {
+		err := fmt.Errorf("negative value of duration")
+		return 0, err
```
gavv, line 48:
```suggestion
		return 0, fmt.Errorf("unexpected negative duration: %v", d)
```

roc/sender.go
```diff
@@ -154,14 +154,20 @@ func OpenSender(context *Context, config SenderConfig) (*Sender, error) {
 		return nil, errors.New("context is closed")
 	}
 
+	cPacketLength, err := go2cUnsignedDuration(config.PacketLength)
+
+	if err != nil {
+		return nil, fmt.Errorf("invalid PacketLength: %w", err)
+	}
```
gavv, line 161:
```suggestion
	cPacketLength, err := go2cUnsignedDuration(config.PacketLength)
	if err != nil {
		return nil, fmt.Errorf("invalid config.PacketLength: %w", err)
	}
```

roc/receiver.go
```diff
@@ -185,19 +186,57 @@ func OpenReceiver(context *Context, config ReceiverConfig) (*Receiver, error) {
...
+		cMaxLatencyOverrun       ulonglong
+		cMaxLatencyUnderrun      ulonglong
+		cBreakageDetectionWindow ulonglong
+	)
+
+	convD := []struct {
+		name string
+		goD  time.Duration
+		cD   *ulonglong
+	}{
```
gavv, line 200: This temporary array and for-loop looks like over-complication for this simple piece of code.

Let's just use separate local variables, similar to cPacketLength in OpenSender().

gavv:
> [Coverage Status](https://coveralls.io/builds/60211925)
> 
> Coverage: 92.687% (-1.4%) from 94.04% when pulling **[0d77f4e](https://github.com/roc-streaming/roc-go/commit/0d77f4e996a33dcc5a0b313ee5483fa3152f99f4) on g41797:main** into **[9647d5d](https://github.com/roc-streaming/roc-go/commit/9647d5d8bb3496fdbdaf49268f72475b24ebcff6) on roc-streaming:main**.

I just notices that the coverage was decreased:

* https://coveralls.io/builds/60211925/source?filename=roc%2Fsender.go
* https://coveralls.io/builds/60211925/source?filename=roc%2Freceiver.go

That's because the newly added error paths are not covered.

We need to add appropriate tests for sender and receiver.

g41797:
1f44d

g41797:
roc/sender_receiver_test.go was deleted (see #97)
but I didn't sync repos
It's the reason of the conflict 
how to proceed?

gavv:
> roc/sender_receiver_test.go was deleted (see https://github.com/roc-streaming/roc-go/pull/97)

It was not deleted, it was renamed to `e2e_test.go`.

gavv:
Note: there were no need to close PR and create a new one in this case; what you need, instead, is to fetch fresh `main` from upstream repo, rebase your forked branch on it, and force-push to your forked repo. PR will be automatically updated after force-push.

g41797:
thanks
saved in "how to ..."

Issue #84, opened 2023-05-13
Use stringer for all enums


gavv:
We have a lot of enums in config.go and log.go, and it would be convenient if they all had String() method that returned enum's symbolic representation. This can be easily achieved using [stringer](https://pkg.go.dev/golang.org/x/tools@v0.9.1/cmd/stringer) tool.

Steps:

* add stringer magic comment to all enums: Interface, Protocol, FecEncoding, PacketEncoding, FrameEncoding, ChannelSet, ResamplerBackend, ResamplerProfile, ClockSource, LogLevel (note: no need to add it to Slot, it's not a enum)

* configure stringer to trim enum's prefix

* add Makefile target that runs go generate and document it in README

* add CI step (in build.yml) checking that we did not forget to run go generate and commit results (it may run go generate on CI and then check git status)

fastbyt3:
@gavv can I try this?

gavv:
Sure, thank you!

gavv:
Landed

PR #98, opened 2023-05-23
 Use stringer for all enums

fastbyt3:
PR for #84 

Finished

- [x] Add stringer comment to all enums (Interface, Protocol, FecEncoding, PacketEncoding, FrameEncoding, ChannelSet, ResamplerBackend, ResamplerProfile, ClockSource, LogLevel)
- [x] use `trimprefix` 
- [x] new Makefile target `generate` and documented in README
- [x] job in CI to check if `go generate` output was commited
- [x] test cases


gavv:
Please check why CI is failing. I think you forgot to commit generated files?

Makefile
```diff
@@ -1,7 +1,10 @@
 GO111MODULE := on
 export GO111MODULE
 
-all: build lint test
+all: generate build lint test
+
+generate:
+	cd roc && go generate
```
gavv, line 7:
```suggestion
all: gen build lint test

gen:
	cd roc && go generate
```

README.md
```diff
@@ -185,7 +185,7 @@ make
 Run only specific checks:
 
 ```
-make build|lint|test
+make generate|build|lint|test
```
gavv, line 188:
```suggestion
make gen|build|lint|test
```

roc/log.go
```diff
@@ -16,6 +16,8 @@ import (
 )
 
 // LogLevel defines the logging verbosity.
+//
+//go:generate stringer -type LogLevel -trimprefix LogLevel
```
gavv, line 20:
```suggestion
//go:generate stringer -type LogLevel -trimprefix Log
```

roc/stringer_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestStringer(t *testing.T) {
+	for i := -10; i <= 1000; i++ {
+		assert.NotEmpty(
+			t,
+			Interface(i).String(),
+			"Interface String func must always return something",
```
gavv, line 14: I think these messages:

```
"Interface String func must always return something"
```

are redundant, because they basically repeat what assert.NotEmpty means. I suggest to remove them.

.github/workflows/build.yaml
```diff
@@ -20,6 +20,34 @@ on:
     - cron: '0 0 * * 1'
 
 jobs:
+  generate:
```
gavv, line 23: nit: let's move it after "formatting" job and rename to "codegen" or "generation"

.github/workflows/build.yaml
```diff
@@ -20,6 +20,34 @@ on:
...
+    runs-on: ubuntu-latest
+    name: check go generate
```
gavv, line 25:
```suggestion
    name: Code generation
```

.github/workflows/build.yaml
```diff
@@ -20,6 +20,34 @@ on:
...
+          cd roc
+          go generate
+
+      - name: Check for git changes
+        run: |
+          git diff --quiet
+          if [[ $? -ne 0 ]]; then
+            echo "go generate might not have been run. Please run it and commit changes"
+            exit 1
+          fi
```
gavv, line 49: Seems this check does not work and needs to be adjusted.

```
$ gh co 98
...
Switched to branch 'stringer_work'

$ make generate 
cd roc && go generate

$ git st
On branch stringer_work
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	roc/channelset_string.go
	roc/clocksource_string.go
	roc/fecencoding_string.go
	roc/frameencoding_string.go
	roc/interface_string.go
	roc/loglevel_string.go
	roc/packetencoding_string.go
	roc/protocol_string.go
	roc/resamplerbackend_string.go
	roc/resamplerprofile_string.go

$ git diff --quiet ; echo $?
0
```

fastbyt3: Yeah i think using `git status --porcelain` would be better

gavv: Yes, and likely we should analyze its output instead of checking exit code

fastbyt3: uh by analyzing output do you mean setting a flag when checking and adding an extra step in the job with if condition

gavv: I thin k we just need a shell command that looks at git status output and fails if it detects modified or untracked files.

gavv: Your current code looks good (if it works; I didn't test it)

```
if [[ $(git status --porcelain | grep -E "[MARCD].*\.go$") ]]; then
    echo "Run go generate and commit generated files"
    exit 1
  fi
  shell: /usr/bin/bash -e {0}
```

gavv:
When it's ready for review, please leave a comment or click "request review"

gavv:
Pushed small fix: https://github.com/roc-streaming/roc-go/pull/98/commits/f30106aef19615e7894f4f0917665412bf7caeb2

gavv:
Thanks!

PR #99, opened 2023-05-24
Use time.Duration for time intervals

g41797:
PR for #83 
Fork was updated before changes

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -19,6 +19,10 @@ on:
   schedule:
     - cron: '0 0 * * 1'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   linux:
     runs-on: ubuntu-latest
--- a/README.md
+++ b/README.md
@@ -200,6 +200,28 @@ Format code:
 make fmt
 ```
 
+## Release
+
+To release a new version:
+
+ * Create git tag
+ 
+    ```
+    ./tag.py --push <remote> <version>
+    ```
+    
+    e.g.
+    
+    ```
+    ./tag.py --push origin 1.2.3
+    ```
+    
+    Or use **tag.py** without **--push** to only create a tag locally, and then push it manually.
+
+* Wait until "Release" CI job completes and creates GitHub release draft.
+
+* Edit GitHub release created by CI and publish it.
+
 ## Authors
 
 See [here](https://github.com/roc-streaming/roc-go/graphs/contributors).
--- a/roc/config.go
+++ b/roc/config.go
@@ -471,7 +471,7 @@ type ReceiverConfig struct {
 	// Target latency, in nanoseconds.
 	// The session will not start playing until it accumulates the requested latency.
 	// Then, if resampler is enabled, the session will adjust its clock to keep actual
-	// latency as close as close as possible to the target latency.
+	// latency as close as possible to the target latency.
 	// If zero, default value is used.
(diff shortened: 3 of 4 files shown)


roc/receiver_test.go
```diff
@@ -35,6 +35,46 @@ func TestReceiver_Open(t *testing.T) {
...
 			wantErr: newNativeErr("roc_receiver_open()", -1),
 		},
+		{
+			name: "invalid config.TargetLatency",
+			contextFunc: func() *Context {
+				ctx, err := OpenContext(makeContextConfig())
+				require.NoError(t, err)
+				return ctx
+			},
+			config:  makeReceiverConfig(WithTargetLatency(-1)),
```
gavv, line 45: Regarding this bit:

```go
makeReceiverConfig(WithTargetLatency(-1))
```

I don't like a few things about this approach:

* test table become less self-contained; we need to look into implementation of functions like WithTargetLatency to see what they're doing

* we need to craft dozens of small functions, one per each field

* we complicate config construction by introducing concept of "options"

It seems there is a way to make things more simple and compact by modifying this approach a bit:

```go
name: "invalid config.TargetLatency",
contextFunc: func() *Context {
	ctx, err := OpenContext(makeContextConfig())
	require.NoError(t, err)
	return ctx
},
configFunc(): func() ReceiverConfig {
	cfg := makeReceiverConfig()
	cfg.TargetLatency = -1
	return cfg
},
```

roc/receiver_test.go
```diff
@@ -35,6 +35,46 @@ func TestReceiver_Open(t *testing.T) {
...
+			wantErr: makeTDError(-1, "config.TargetLatency"),
```
gavv, line 46: I think a small portion of copy-paste won't harm in this case and we can avoid introducing extra helper (makeTDError) to keep test table more descriptive:

```
wantErr: fmt.Errorf("invalid config.TargetLatency: %w",
		fmt.Errorf("unexpected negative duration: -1")
```

Like we do in all other tests.

gavv (commented):
Thanks for update! A few more comments.

roc/convert.go
```diff
@@ -37,3 +41,10 @@ func go2cBool(b bool) C.uint {
 	}
 	return 0
 }
+
+func go2cUnsignedDuration(d time.Duration) (ulonglong, error) {
```
gavv, line 45: Please add unit test for this func to `conver_test.go`

roc/receiver_test.go
g41797: no problem
after fix i'll update you via re-request review

roc/receiver_test.go
g41797: got the point

g41797:
problems with environment
fix will take sometime

gavv:
> problems with environment - ./receiver.go:313:13: could not determine kind of name for C.roc_receiver_set_reuseaddr
> 
> dejavu - in the past I solved this problem but don't remember how :-(
> 
> do you have some advise?
> 
> I will use re-request review to raise your attention

What version of libroc do you have installed on system? You can find it in `/usr/include/roc/version.h` or `/usr/local/include/roc/version.h`.

You can try installing latest version from `master` branch.

g41797:
/usr/include/roc/version.h - 0.2.1 (github - 0.2.4 ) receiver.h doesn't include _roc_receiver_set_reuseaddr_

/usr/local/include - empty 

as far as I remember was installed from https://roc-streaming.org

after installation using zip from github version.h - 0.2.4

compilation - OK

Thanks a lot

PR #109, opened 2023-06-04
Use time.Duration for time intervals

g41797:
Implementation of #83


gavv:
Thanks for PR. Here is a small follow-up commit: c4ec4019003bf1609497ec713bc549f125ab82e7

PS. please don't recreate PRs, its hard to track review across multiple requests. Instead you can always force-push your changes to PR's branch.

PR #111, opened 2023-07-02
improve e2e test

### LINKED ISSUE #101: Improve acceptance criteria of end-to-end test
TestEnd2End_Default is a simple test that works as follows:

* creates sender and receiver, and connects sender to receiver
* in the first goroutine, repeatedly writes samples to sender
* in the second goroutine, repeatedly reads samples from receiver
* waits until it receives two consecutive non-zero samples and exits

This algorithm is the most basic check that verifies that if we send "something", then we eventually receiver "something", with a very relaxed definition of "something" (two non-zero samples).

----

It was implemented like this because we can't implement very strict check: we can't rely that we receive the exact same stream as we send.

When resampler and other processing is disabled, a few differences are possible between sent and received streams:

* received stream can produce arbitrary amount of zero samples before it starts producing the original stream
* arbitrary chunks of original stream may be replaced with zeros in received stream
* frame boundaries are not preserved; the first frame written to the sender may start in the middle of the first non-zero frame read from receiver

----

Keeping this in mind, we can improve TestEnd2End: make the check more strict and meaningful than it is currently, though relaxed enough to take into account possible stream modifications described above.

Here is an algorithm that we could implement in the test:

* on sender, produce a repeating sequence of incrementing numbers `0 1 2 ... 99 100 0 1 2 ...`; we can use this sequency for L channel and the same but negated sequence for R channel

* on receiver:

    * wait until we receive first non-zero sample

    * check that since then, each sample either have expected value (so that the next sample is previous sample plus one), or is zero (if corresponding packet was lost)

    * also check that either L = -R or L = R = 0

    * wait until we accumulate 10'000 non-zero samples in total (not taking into account possible zero samples)

* both on sender and receiver, we should limit the speed of invocation of WriteFloats and ReadFloats methods

    For example, if sample rate in sender/receiver config is 44100, we should try to write and read only 44100 samples per seconds; e.g., if our frame size is 100, then we should write frame each 1/441 = 0.002267 seconds.

    We can create tickers for both sender and receiver with corresponding frequency and wait a message from ticker before invoking write or read next time.

ortex:
fixes #101

### DIFF
--- a/roc/e2e_test.go
+++ b/roc/e2e_test.go
@@ -1,7 +1,9 @@
 package roc
 
 import (
+	"fmt"
 	"math"
+	"strings"
 	"sync"
 	"testing"
 	"time"
@@ -9,6 +11,8 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
+
+const NumChannels = 2
 
 type e2e struct {
 	Context  *Context
@@ -86,10 +90,10 @@ func TestEnd2End_Default(t *testing.T) {
 	defer e.close(t)
 
 	samplesCnt := 100
-	samples := make([]float32, samplesCnt)
-	for i := 0; i < samplesCnt/2; i++ {
-		samples[i*2] = float32(i+1) / 100
-		samples[i*2+1] = -float32(i+1) / 100
+	testSamples := make([]float32, samplesCnt)
+	for i := 0; i < samplesCnt/NumChannels; i++ {
+		testSamples[i*NumChannels] = float32(i+1) / 100
+		testSamples[i*NumChannels+1] = -float32(i+1) / 100
 	}
 
 	var interval = time.Second / time.Duration(44100/samplesCnt)
@@ -106,7 +110,7 @@ func TestEnd2End_Default(t *testing.T) {
 		for {
 			select {
 			case <-sendTicker.C:
-				err := e.Sender.WriteFloats(samples)
+				err := e.Sender.WriteFloats(testSamples)
 				if err != nil {
 					t.Fail()
 				}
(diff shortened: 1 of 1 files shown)


gavv:
Nice!

I few suggestions:

* Test waits until it receives first non-zero sample, then receives 10000 samples, then checks that each one is either zero or has correct value; it means that if we received one non-zero sample, and then 10000 zeros, the test will pass. Instead of this, I suggest to remove upper limit for how much samples we receive, and instead place lower limit of how much non-zero samples we want. So we will receive samples, check that each one is either zero or correct, and repeat this until we accumulate N non-zero samples. This way the test will check that there are a lot (N) of good samples, and at the same time it won't be bound to a particular loss ratio. Losses may happen on a loaded CI worker.

* 100 samples is not very interesting number - they will fit in one packet. I suggest to wait until we receive something like 10K non-zero samples, so that it will be many packets.

* To avoid unintuitive part with 0.5, we can work with integers:

   ```
   valueL = int(samples[i*2] * 100)
   assert((prevValue + 1) % 100 == valueL)
   ```

* When assertions fails, it makes sense to report index and maybe dump the whole stream or block.

* Let's replace magical number `2` with a constant like `NumChan(nel)s`.

ortex:
> * When assertions fails, it makes sense to report index and maybe dump the whole stream or block.

```
--- FAIL: TestEnd2End_Default (0.42s)
    e2e_test.go:146: 
                Error Trace:    /Users/andreybushmin/workspace/roc/roc-go/roc/e2e_test.go:146
                Error:          Not equal: 
                                expected: 2
                                actual  : 3
                Test:           TestEnd2End_Default
                Messages:       prevL: 1, valueL: 2, index: 2, samples: [0=0.01, 1=-0.01, 2=0.02, 3=-0.02, 4=0.03, 5=-0.03, 6=0.04, 7=-0.04, 8=0.05, 9=-0.05, 10=0.06, 11=-0.06, 12=0.07, 13=-0.07, 14=0.08, 15=-0.08, 16=0.09, 17=-0.09, 18=0.10, 19=-0.10, 20=0.11, 21=-0.11, 22=0.12, 23=-0.12, 24=0.13, 25=-0.13, 26=0.14, 27=-0.14, 28=0.15, 29=-0.15, 30=0.16, 31=-0.16, 32=0.17, 33=-0.17, 34=0.18, 35=-0.18, 36=0.19, 37=-0.19, 38=0.20, 39=-0.20, 40=0.21, 41=-0.21, 42=0.22, 43=-0.22, 44=0.23, 45=-0.23, 46=0.24, 47=-0.24, 48=0.25, 49=-0.25, 50=0.26, 51=-0.26, 52=0.27, 53=-0.27, 54=0.28, 55=-0.28, 56=0.29, 57=-0.29, 58=0.30, 59=-0.30, 60=0.31, 61=-0.31, 62=0.32, 63=-0.32, 64=0.33, 65=-0.33, 66=0.34, 67=-0.34, 68=0.35, 69=-0.35, 70=0.36, 71=-0.36, 72=0.37, 73=-0.37, 74=0.38, 75=-0.38, 76=0.39, 77=-0.39, 78=0.40, 79=-0.40, 80=0.41, 81=-0.41, 82=0.42, 83=-0.42, 84=0.43, 85=-0.43, 86=0.44, 87=-0.44, 88=0.45, 89=-0.45, 90=0.46, 91=-0.46, 92=0.47, 93=-0.47, 94=0.48, 95=-0.48, 96=0.49, 97=-0.49, 98=0.50, 99=-0.50]
```

output example (changed assert to fail)

all other suggestions are fixed

gavv:
Thanks!!

Issue #104, opened 2023-05-29
Add end-to-end test with internal clock


gavv:
TestEnd2End_Default is a simple integration test that writes samples to sender, reads them from receiver, and verifies the received stream.

TestEnd2End_Default sets ClockSource to ClockExternal both in sender and receiver config.

We should add a new test TestEnd2End_Blocking that sets ClockSource to ClockInternal both for sender and receiver.

TestEnd2End_Default and TestEnd2End_Blocking should share most of the code, so that the algorithm described in #101 would be used for both of them. However, in TestEnd2End_Blocking, we should disable tickers described in #101.

For further details about ClockExternal vs ClockInternal, see #100.

Issue #102, opened 2023-05-29
Add end-to-end test without FEC


gavv:
TestEnd2End_Default is a simple integration test that writes samples to sender, reads them from receiver, and verifies the received stream.

Currently TestEnd2End_Default enables Reed-Solomon forward error correction (FEC) by setting SenderConfig.FecEncoding and binding two endpoints both on sender and receiver: one endpoint for "source" packets (i.e. media) and one for "repair" packets (for error correction).

We need to split it into two tests:

* TestEnd2End_Default - should not use FEC
* TestEnd2End_FEC - should use FEC

The new TestEnd2End_Default should disable FEC in sender config and use only a single endpoint `rtp://127.0.0.1:0` for source packets, and don't use repair endpoint.

TestEnd2End_Default and TestEnd2End_FEC should share most of the code, so that the algorithm described in #101 would be used for both of them.

Related:

* https://roc-streaming.org/toolkit/docs/api.html
* https://roc-streaming.org/toolkit/docs/internals/fec.html

PR #112, opened 2023-08-09
more e2e tests

ortex:
add e2e TestEnd2End_Blocking for ClockInternal clock source; split default e2e test for with/without fec

#102 #104

### DIFF
--- a/roc/e2e_test.go
+++ b/roc/e2e_test.go
@@ -15,9 +15,11 @@ import (
 const NumChannels = 2
 
 type e2e struct {
-	Context  *Context
-	Receiver *Receiver
-	Sender   *Sender
+	ReceiverConfig ReceiverConfig
+	SenderConfig   SenderConfig
+	Context        *Context
+	Receiver       *Receiver
+	Sender         *Sender
 }
 
 type e2eParams struct {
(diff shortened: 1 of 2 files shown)


roc/e2e_test.go
```diff
@@ -86,78 +103,181 @@ func (e *e2e) close(t *testing.T) {
...
-					t.Fail()
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			e := newE2E(t, tt.params)
+			defer e.close(t)
+
+			samplesCnt := 100
+			testSamples := generateTestSamples(samplesCnt)
+
+			var interval = time.Second / time.Duration(44100/samplesCnt)
```
gavv, line 135: nit: Let's use sample rate from config, instead of 44100?

gavv (commented):
Thanks! One small comment, not even related to PR itself.

roc/e2e_test.go
ortex: done!

ortex:
@gavv Could you have a look?

fixed the discussion and set go-version to 1.20 instead of the latest because our ci pipeline fails with new go 1.21, see https://github.com/roc-streaming/roc-go/actions/runs/6067814352 I will create a separate issue for that

Btw, need to use '1.20' as go version for action instead of 1.20, because 1.20 is treated as 1.2 https://github.com/actions/setup-go/issues/326#issuecomment-1561222838

gavv:
Thanks! Damn, it was float!

Issue #108, opened 2023-06-02
Implement "-f" option in tag.py script


gavv:
`tag.py` is a script for creating and pushing git tag (#106).

If the tag already exists locally or remotely, it will fail, which is usually OK, but sometimes we need to override recently created malformed tag.

It would be nice to add an option called `-f` / `--force`, which allows to override existing tag, both local and remote (in case when `--push` is specified).

gavv:
Landed, thanks @pypaut!

PR #114, opened 2023-09-08
Add --force option to tag script (#108)

pypaut:
This is my first ever pull request to an open-source repository, please do not hesitate to give feedback.

### DIFF
--- a/tag.py
+++ b/tag.py
@@ -28,9 +28,15 @@ def run_command(args):
     print(f'--- Running command: {" ".join(args)}')
     subprocess.check_call(args)
 
-def commit_change(version):
+def commit_change(version, force):
     run_command(['git', 'add', 'roc/version.go'])
-    run_command(['git', 'commit', '-m', f'Release {version}'])
+    try:
+        run_command(['git', 'commit', '-m', f'Release {version}'])
+    except subprocess.CalledProcessError:
+        if force:
+            print('--- Nothing to commit, but --force was used')
+        else:
+            raise
 
 def create_tag(tag, force):
     if force:
@@ -74,7 +80,7 @@ def main():
         sys.exit(1)
 
     write_version(version)
-    commit_change(version)
+    commit_change(version, force)
     create_tag(tag, force)
     if remote:
         push_change(remote, tag, force)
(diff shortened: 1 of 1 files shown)


ortex:
Hi, thanks for PR!

Could you please attach the console log or/and screenshots of how you tested it?

pypaut:
Hello, I had tested on a different repository, did it again more thoroughly on roc-go.
Here are my logs: https://privatebin.net/?16578bd480ae82e8#EgzUM9TJKCXdSByAyx1bv9duYcrAbKnLjK3TBDcf9pff

tag.py
```diff
@@ -28,23 +28,38 @@ def run_command(args):
     print(f'--- Running command: {" ".join(args)}')
     subprocess.check_call(args)
 
-def commit_change(version):
+def commit_change(version, force):
     run_command(['git', 'add', 'roc/version.go'])
-    run_command(['git', 'commit', '-m', f'Release {version}'])
+    try:
+        run_command(['git', 'commit', '-m', f'Release {version}'])
+    except subprocess.CalledProcessError:
```
ortex, line 35: I think it's better to check if we have something to commit instead of catching a general error here

pypaut: Done! It feels better this way, thank you.

I did the same tests as in the previous privatebin.

gavv: Small suggestion: we could first run "git add", and only then "git diff", but with "--cached" option. This way we will handle all 3 situations: when roc/version.go is unchanged, changed and not staged, or changed and staged.

gavv (approved):

tag.py
```diff
@@ -28,23 +28,39 @@ def run_command(args):
     print(f'--- Running command: {" ".join(args)}')
     subprocess.check_call(args)
 
-def commit_change(version):
+def check_version_to_commit():
+    cmd = ['git', 'diff', '--cached', 'roc/version.go']
+    return subprocess.check_output(cmd) != b''
+
+def commit_change(version, force):
```
ortex, line 35: nit: `force` arg can be removed now

ortex (approved):

ortex:
Thanks @pypaut !

We have a similar script in another repo https://github.com/roc-streaming/roc-droid/blob/main/tag.py it would be nice to update it as well 😄

